Issue #72, opened 2015-05-11
Country name in locale selector should use the target language


Ken-Lo:
Country names in the locale selector shouldn't be translated (they should always appear the language of the target page).

For instance, with the current setup (translated country names), someone on a Greek-language page would need to choose "Ηνωμένο Βασίλειο" to go back to English/UK (bad usability).

**Current Setup:**

![country selector fr](https://cloud.githubusercontent.com/assets/6426688/7567749/a2e872b4-f7f6-11e4-816f-f474d97db9b4.png)

**Intended Behaviour:**

Country name provided in the language of the target page. e.g.

![ryanair language switcher](https://cloud.githubusercontent.com/assets/6426688/7472431/7f9741b2-f32a-11e4-88e4-d2ed1546cf28.png)

Probably requires some new code to grab the country names out of the `app/messages/` folder?

PR #147, opened 2015-05-20
Show couuntry name in target lang

harrison:
Shows country name in country selector in the target language.

Fixes: https://github.com/gocardless/splash-pages/issues/72

### DIFF
--- a/app/components/app/app.js
+++ b/app/components/app/app.js
@@ -32,30 +32,32 @@ function setCanonicalHref(props) {
 class App extends React.Component {
   displayName = 'App'
 
+  static propTypes = {
+    locales: PropTypes.locale,
+    messages: PropTypes.object.isRequired,
+    formats: PropTypes.object.isRequired,
+    routeName: PropTypes.string.isRequired,
+    availableLocales: PropTypes.array.isRequired,
+    availableCountryNames: PropTypes.object.isRequired,
+    routeLocales: PropTypes.array.isRequired,
+    pathname: PropTypes.string.isRequired,
+    config: PropTypes.object.isRequired,
+  }
+
   static contextTypes = {
-    router: React.PropTypes.func,
+    router: PropTypes.func,
   }
 
   static childContextTypes = {
     locales: PropTypes.locale,
-    messages: React.PropTypes.object.isRequired,
-    formats: React.PropTypes.object.isRequired,
-    routeName: React.PropTypes.string.isRequired,
-    availableLocales: React.PropTypes.array.isRequired,
-    routeLocales: React.PropTypes.array.isRequired,
-    pathname: React.PropTypes.string.isRequired,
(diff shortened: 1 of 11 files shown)


harrison:
Review @english?

app/components/app/app.js
```diff
@@ -32,30 +32,32 @@ function setCanonicalHref(props) {
 class App extends React.Component {
   displayName = 'App'
 
-  static contextTypes = {
-    router: React.PropTypes.func,
+  static propTypes = {
```
english: The only change to this file is adding `availableCountryNames` as a `prop` and making it available in its `context`? The diff is really hard to read!

app/components/footer/footer.js
```diff
@@ -9,19 +9,18 @@ import Translation from '../translation/translation';
...
 
-function buildAvailableLocalePages(currentLocale, siteLocales, messages) {
+function buildAvailableLocalePages(currentLocale, siteLocales, availableCountryNames) {
   const availableLocales = Object.keys(siteLocales).map(function(locale) {
     const isActive = currentLocale === locale;
     const path = siteLocales[locale].path;
-    const name = getMessage(messages, `country_names.${locale}`);
-    return {locale, isActive, path, name};
+    const name = availableCountryNames[locale];
+    return {locale, isActive, path, name };
```
english: Do you know why `eslint` didn't pick up on the spacing here?

app/components/footer/footer.js
```diff
@@ -33,17 +32,17 @@ class Footer extends React.Component {
...
   }
 
   render() {
-    const { routeName, locales, messages, config, availableLocales } = this.context;
+    const { routeName, locales, config, availableLocales, availableCountryNames } = this.context;
 
     const siteLocales = merge({}, getLocalesForRouteName(homeRoute, availableLocales),
       getLocalesForRouteName(routeName, availableLocales));
-    const availableLocalePages = buildAvailableLocalePages(locales, siteLocales, messages);
+    const availableLocalePages = buildAvailableLocalePages(locales, siteLocales, availableCountryNames);
```
english: Could the `siteLocales` assignment get moved into `buildAvailableLocalePages`? Might make it a bit easier to follow, it seems its easy for the `render` methods to get quite complex. Do you agree?

app/components/app/app.js
harrison: yeah sorry about that, the order was weird before

english:
Looks good!

Its a shame there's so much code required in order to pass down context from top to bottom.

Am I right in thinking that, with this, the country switcher will only show flags for countries that have corresponding pages for the current path?

harrison:
Yeah, we might be able to have one context wrapper class that both the html document and app can use, will see if that works

harrison:
The country switcher will show all available countries for the site, if a page doesn't exist it links to the home page for that locale

Issue #143, opened 2015-05-20
Splash pages give "Not found" error in Google cache


Ken-Lo:
All splash pages give a "Not found" error when viewed in Google's cache with Javascript enabled. With Javascript disabled, it works OK. Pages seem to be crawled correctly (so a fairly low priority fix).

![screen shot 2015-05-20 at 14 59 44](https://cloud.githubusercontent.com/assets/6426688/7727486/e198df4c-ff00-11e4-85e0-507bd8e309b4.png)
http://webcache.googleusercontent.com/search?q=cache:Z60paZAQ7wsJ:https://gocardless.com/+&cd=1&hl=en&ct=clnk&gl=uk

I suspect it's probably a weird conflict in the Javascript as the page is being accessed through path `/search?q=cache:Z60paZAQ7wsJ:https://gocardless.com/+&cd=1&hl=en&ct=clnk&gl=uk` (possibly, also an issue with cross-domain Javascript).

Probably not a major issue as it doesn't affect indexing (only seems to affect people viewing our website through a weird HTTP proxy, etc). Can probably be fixed by checking the hostname in bundle js? Though feel free to mark as Won't Fix if it's difficult to do or if it causes problems elsewhere.

harrison:
Ugh this is annoying, will definitely be the query string. When js is enabled it will try and navigate to `/search`.. also the 404 page is really shit, opening an issue for this

PR #148, opened 2015-05-20
Whitelist allowed origins running js version of the site

harrison:
Probably a terrible idea.. need think this through.

### DIFF
--- a/config/blacklisted-origins.js
+++ b/config/blacklisted-origins.js
@@ -1,0 +1,9 @@
+import includes from 'lodash/collection/includes';
+
+export const blacklistedHosts = [
+  'webcache.googleusercontent.com',
+];
+
+export function isHostBlacklisted(host) {
+  return includes(blacklistedHosts, host);
+}
--- a/app/client/bootstrap.js
+++ b/app/client/bootstrap.js
@@ -5,10 +5,10 @@ import findLast from 'lodash/collection/findLast';
 import assign from 'lodash/object/assign';
 import React from 'react';
 import Router from 'react-router';
-import {getRoutes} from '../router/route-helpers';
-import {pushDataLayer} from '../helpers/gtm-tracker/gtm-tracker';
+import { getRoutes } from '../router/route-helpers';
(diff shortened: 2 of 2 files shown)


harrison:
Fixes https://github.com/gocardless/splash-pages/issues/143

english:
What's the state of this @harrison?

harrison:
Good question, what do you reckon about the approach?:) feels icky..

english:
Yeah, I see what you mean. There's not many alternatives.

What about just checking whether we're in an iframe? Something along the lines of: http://stackoverflow.com/questions/326069/how-to-identify-if-a-webpage-is-being-loaded-inside-an-iframe-or-directly-into-t

Ken-Lo:
@english I'm not sure it's technically an iframe, Google cache simply outputs the HTML of our page and  then calls cross-domain Javascript from http://webcache.googleusercontent.com/. So not sure the code there will work.

All-in-all, it's not super urgent to fix this issue especially if it's icky or has the chance of breaking things. Blacklisting googleusercontent.com, web.archive.org, etc. is probably another way to do this - equally as icky but probably has less chance of breaking things.

english:
Thanks @Ken-Lo, I didn't realise that. In that case, then I guess this either blacklisting or whitelisting would be fine.

harrison:
hm yeah might do a blacklist instead

harrison:
@english wanna give this a last check? webarchive doesn't run any js so no need to blacklist

config/blacklisted-origins.js
```diff
@@ -0,0 +1,7 @@
+export const blacklistedHosts = [
+  'webcache.googleusercontent.com',
+];
+
+export function isHostBlacklisted(host) {
+  return blacklistedHosts.some((blanklistedHost) => blanklistedHost === host);
```
english: I'm curious if there's any reason you didn't use `_.contains` or `indexOf`?

english:
Looks good to me :+1:

config/blacklisted-origins.js
harrison: oh yeah good point, was doing a regex before, so needed `some`

PR #149, opened 2015-05-20
Watch node land

harrison:
Temp fix to reload the node server when messages changes. Still need to reload the browser.

Want to figure out a better way to deal with messages, maybe write a webpack plugin..


harrison:
@english review?

english:
:+1:

PR #151, opened 2015-05-21
Updating outdated and node 0.12

harrison:


package.json
```diff
@@ -16,53 +16,53 @@
...
-    "webpack": "1.9.5",
-    "webpack-dev-server": "1.8.2"
+    "serve": "^1.4.0",
+    "serve-favicon": "^2.2.1",
+    "strip-loader": "^0.1.0",
+    "style-loader": "^0.12.2",
+    "superagent": "^1.2.0",
+    "touch": "^0.0.3",
+    "webpack": "^1.9.7",
+    "webpack-dev-server": "^1.8.2"
```
jackfranklin: @harrison we purposefully swapped from using the `^` to just hard coding an actual version number so we could keep a stricter lid on things. Any reason not to do that?

harrison: oh but it doesn't really make any difference with the shrinkwrap, also this is what node writes out when you do `npm install thing@0.1`

PR #154, opened 2015-05-21
Fix 404 pages

harrison:
screen shot 2015-05-21 at 11 26 55

### DIFF
--- a/app/pages/not-found/not-found.js
+++ b/app/pages/not-found/not-found.js
@@ -1,14 +1,29 @@
 import React from 'react';
-import Page from '../../components/page/page';
+import Message from '../../components/message/message';
 
 export default class NotFound extends React.Component {
   displayName = 'NotFound'
 
   render() {
     return (
-      <Page>
-        <h1>Not found</h1>
-      </Page>
+      <div className='not-found'>
+        <div className='not-found__container'>
+          <div className='not-found__container__center'>
+            <div className='not-found__container__center__inner'>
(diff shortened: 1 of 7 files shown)


harrison:
hm this is pretty ghetto.. this is only when hitting a 404 page in JS, which should pretty much never happen. In prod the only 404 page is the one set by s3, so you can never get a translated 404 page.. only way to really deal with this is run the express server in prod:(

@english thoughts?

app/pages/not-found/not-found.js
```diff
@@ -1,14 +1,33 @@
 import React from 'react';
-import Page from '../../components/page/page';
+import Message from '../../components/message/message';
+
+if (process.env.BROWSER) {
+  require('./not-found.css');
```
english: What's going on here?

english: Is it not bundled up with the rest of the app?

harrison: thought it wasn't but is, moving to main

english:
Hmmm. That's a good point about only being able to show English 404s without the web server. One to think about, but it doesn't seem like a such a big deal...

english:
What exactly is the scenario for hitting a JS 404?

harrison:
In development and when loading the site in an iframe or proxied where the browser url is something like `googlecache.com?gocardless.com`

english:
Ah ok, cool. :+1:

PR #156, opened 2015-05-21
Link to blog post

english:
Adds link to new blog post.

Reduces header size since GitHub headers are huge!


english:
@harrison FYI, but doesn't require review.

PR #158, opened 2015-05-21
Fix learn more

### LINKED ISSUE #153: `learn more about Pro` should probably be right-aligned?
On /features it currently looks like this:

screen shot 2015-05-21 at 11 30 38

Pretty sure the link should be aligned with the right hand column in the table, no?

Code looks like this:
```
<td />
<td />
<td className='u-padding-As'>
  <a href='/pro' id='track-comparison-table-pro' className='u-color-primary'>
    Learn more about GoCardless Pro
  </a>
</td>
```

So it's already the third cell across in the table, but the problem is that the first 2 cells in the table basically have zero width.

I'm a bit of a noob to this repo - probably just quicker / easier for someone else to fix? Thanks guys!

english:
Fixes #153

In addition to fixing the alignment, also fixes a place we were using an `a` link, rather than a `Link` component, meaning you would always end up on the English Pro page.

Before:
screen shot 2015-05-21 at 17 29 00

After:
screen shot 2015-05-21 at 16 57 07

@harrison could you review?

### DIFF
--- a/app/pages/features/features.en.js
+++ b/app/pages/features/features.en.js
@@ -331,9 +331,9 @@ export default class FeaturesEn extends React.Component {
                       <td />
                       <td />
                       <td className='u-padding-As u-size-1of3'>
-                        <a href='/pro' id='track-comparison-table-pro' className='u-color-primary'>
+                        <Link to='/pro' id='track-comparison-table-pro' className='u-color-primary'>
                           Learn more about GoCardless Pro
-                        </a>
+                        </Link>
                       </td>
                     </tr>
                   </tfoot>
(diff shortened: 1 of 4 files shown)


harrison:
:+1:

Ken-Lo:
Nice!
@english Any reason why we don't move this into the `<ProductComparison />` component (either with a new attribute where you specify the current route & use logic for the links or auto-detect it from parent if possible)? Not sure if it's valid putting the `<tfoot>` in a `<table>` of itself?

samjewell:
@english @Ken-Lo @harrison thanks for fixing, great stuff :sparkle:

english:
Thanks for spotting @samjewell!

@Ken-Lo That's a good point. This PR was just a quick fix for the immediate issue, but I agree there is a fair bit of structural duplication across the site between different locales. @harrison and I are thinking about the best way to do this. Its tough to get the best mix of conciseness and flexibility.

PR #165, opened 2015-05-22
Proof of concept for inlining default message

english:
I don't want to merge this, its here only for discussion.

This explores writing `Message` components that are potentially translated into several languages with a 'default' language inline. This is different to what we had before as there would be no inline copy, all copy would live in a translation file.

The benefit of this is that you can see what the actual page looks like with copy in it as you are reading/writing a React page. Previously you would have to jump between the component and the translation file to get an idea of how the page would look.

@harrison what do you think?

### DIFF
--- a/app/components/inline-message/inline-message.js
+++ b/app/components/inline-message/inline-message.js
@@ -1,0 +1,62 @@
+import React from 'react';
+import { PropTypes } from '../../helpers/prop-types/prop-types';
+import isPlainObject from 'lodash/lang/isPlainObject';
+
+function getMessageFromTranslations(pointer, messages) {
+  return pointer.split('.').reduce(function(obj, pathPart) {
+    return isPlainObject(obj) && obj[pathPart];
+  }, messages);
+}
+
+function isSubsetOfLocale(string, locale) {
+  return locale.startsWith(string);
+}
+
+function checkForDuplicateMessage(localeA, localeB, pointer, messages) {
+  if (localeA !== localeB) { return; }
+
+  const messageExists = !!getMessageFromTranslations(pointer, messages);
+
+  if (messageExists) {
+    throw new Error(`Duplicate translation found for ${pointer}`);
+  }
+}
+
+function getMessage({ currentLocale, locale, givenContent, pointer, messages }) {
+  checkForDuplicateMessage(currentLocale, locale, pointer, messages);
+
+  const foundMessage = getMessageFromTranslations(pointer, messages);
+
+  if (foundMessage) {
+    return foundMessage;
+  }
+
+  if (isSubsetOfLocale(locale, currentLocale)) {
+    return givenContent;
+  }
+
+  throw new ReferenceError(`Could not find Intl pointer: ${pointer}`);
+}
+
+export default class InlineMessage extends React.Component {
+  displayName = 'InlineMessage'
+
+  static propTypes = {
+    children: PropTypes.string.isRequired,
+    locale: PropTypes.string.isRequired,
+    pointer: PropTypes.string.isRequired,
+  }
+
+  static contextTypes = {
+    messages: PropTypes.object.isRequired,
+    currentLocale: PropTypes.locale,
+  }
+
+  render() {
+    const { currentLocale, messages } = this.context;
+    const { locale, children, pointer } = this.props;
+    const message = getMessage({ currentLocale, locale, givenContent: children, pointer, messages });
+
+    return <span>{message}</span>;
+  }
+}
--- a/app/components/inline-message/inline-message.spec.js
+++ b/app/components/inline-message/inline-message.spec.js
@@ -1,0 +1,127 @@
+import InlineMessage from './inline-message';
+import React from 'react';
+import ReactTestUtils from 'react/lib/ReactTestUtils';
+import { PropTypes } from '../../helpers/prop-types/prop-types';
+
+function inlineMessageWithContext(context) {
+  return class extends React.Component {
+    static childContextTypes = {
+      messages: PropTypes.object.isRequired,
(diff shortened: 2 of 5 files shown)


app/components/inline-message/inline-message.js
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import { PropTypes } from '../../helpers/prop-types/prop-types';
+import isString from 'lodash/lang/isString';
+import isPlainObject from 'lodash/lang/isPlainObject';
+
+function getMessageFromTranslations(name, messages) {
+  return name.split('.').reduce(function(obj, pathPart) {
+    return obj && obj[pathPart];
```
harrison: won't this blow up if the path doesn't exist in messages?

harrison: worth re-using getMessage?

english: I've got a test for when the path doesn't exist

english: I suppose an `isPlainObject` test would be more robust.

english: > worth re-using getMessage?

Yeah, I wanted to write it without and then refactor what was duplicated.

app/components/inline-message/inline-message.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  throw new ReferenceError(`Could not find Intl name: ${name}`);
+}
+
+export default class InlineMessage extends React.Component {
+  displayName = 'InlineMessage'
+
+  static propTypes = {
+    children: PropTypes.node.isRequired,
```
harrison: `PropTypes.string.isRequired`?

harrison: don't think you will be able to have a node in the messages file

app/components/inline-message/inline-message.js
```diff
@@ -0,0 +1,63 @@
...
+    givenLocale: PropTypes.string.isRequired,
+    name: PropTypes.string.isRequired,
```
harrison: any reason `name` isn't `message` and `givenLocale` `locale`?

english: I found `message` really confusing, given its a pointer to an actual message. It would ideally be `key` but React already uses that :(

english: I can see an argument for `locale`, I had that initially. It was just a bit confusing when using it in the component since `locales` is already in `context`.

harrison: yeah see your point, should we re-name the prop on `Message` to `name`? would be nice to keep it in sync

harrison: yeah would be nice to change `locales` to `currentLocale` :) thinking that `locale` is easier to remember here on the interface side

english:
I've tidied this up and put some tests around it. @harrison were you happy with your review? The rename of `context.locales` => `context.currentLocale` is in a separate PR.

app/components/inline-message/inline-message.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import { PropTypes } from '../../helpers/prop-types/prop-types';
+import isPlainObject from 'lodash/lang/isPlainObject';
+
+function getMessageFromTranslations(name, messages) {
```
english: @harrison not sure how to refactor this one.

`getMessage` throws an exception if it can't find one. I could reuse that one and catch the exception, but in JS that's a bit shit.

Alternatively, I could extract the common bit out and call it something different, but I can't think of a good name! In Ruby I'd just put a bang on the end...

app/components/inline-message/inline-message.js
english: Regarding renaming the `message` prop of `Message`, we also use `message` for the same purpose on `Link` and `Href`. Renaming it to `name` makes sense everywhere but `Link` where `name` could be confused with the `a[name]` attribute. Can you think of a better one :wink:

app/components/inline-message/inline-message.js
harrison: yeah I see, maybe not worth it then

app/components/inline-message/inline-message.js
```diff
@@ -0,0 +1,62 @@
...
+  throw new ReferenceError(`Could not find Intl name: ${name}`);
+}
+
+export default class InlineMessage extends React.Component {
+  displayName = 'InlineMessage'
+
+  static propTypes = {
+    children: PropTypes.string.isRequired,
+    givenLocale: PropTypes.string.isRequired,
+    name: PropTypes.string.isRequired,
```
harrison: hm what about `path`?

harrison: oh wait that doesn't work for Link..

harrison: really against `message`? :) would rather keep it the same everywhere

harrison: or `pointer`

app/components/inline-message/inline-message.js
```diff
@@ -0,0 +1,62 @@
...
+
+  throw new ReferenceError(`Could not find Intl name: ${name}`);
+}
+
+export default class InlineMessage extends React.Component {
+  displayName = 'InlineMessage'
+
+  static propTypes = {
+    children: PropTypes.string.isRequired,
+    givenLocale: PropTypes.string.isRequired,
```
harrison: `locale`?

harrison:
Looking great! Minor comments

app/components/inline-message/inline-message.js
english: I'll go with `pointer`. Its a bit vague but doesn't clash with any other concept!

PR #169, opened 2015-05-26
Rename context.locales => context.currentLocale

english:
Renames `locales` to `currentLocale`. `locales` is confusing as there is only ever one.

This value is typically passed down through our component's `context`, so is therefore referenced almost everywhere! Tests seem fine and I've clicked around locally and everything seems fine...

@harrison could you please review?


harrison:
Awesome! :+1:

PR #170, opened 2015-05-27
Rename message => pointer

english:
This is used to lookup the correct translation. Message was a bit confusing since it
wasn't the actual message, it was just a `key` to it. `key` is already used by
`React.Component`, so `pointer` is the next best thing.

@harrison look good?


harrison:
:+1: also in the readme

PR #173, opened 2015-05-28
Use span rather than div for link exists to avoid CSS problems

jackfranklin:
@harrison is this the right approach? Swapping the wrapping element to be one that isn't block by default makes sense to me...

Good spot from @jesssummerfield  !

Before
screen shot 2015-05-28 at 17 46 13

After
screen shot 2015-05-28 at 18 08 08

### DIFF
--- a/app/components/link-exists/link-exists.js
+++ b/app/components/link-exists/link-exists.js
@@ -24,8 +24,10 @@ class LinkExists extends React.Component {
       hasLink = false;
     }
 
+    if (!hasLink) { return null; }
+
     return (
-      <div>{ hasLink && this.props.children }</div>
+      <span>{ this.props.children }</span>
     );
   }
 }


harrison:
:+1:

Ken-Lo:
Nice one, cheers @jackfranklin (sorry, didn't notice how it affected other pages).

I assume React.js only works if you actually surround the element with some type of tag? I'd have thought it would technically be [invalid HTML](https://docs.webplatform.org/wiki/html/elements/ul) to have a `<span>` underneath `<ul>`? Also to end up with a block `<li>` element in an inline `<span>`, though not sure if this actually affects rendering in any engines.

Wondering if it's possible to do something like this in React?
````jsx
<li IfLinkExists="jobs">
   Jobs
</li>
````
That way, we can enforce a strict relationship between the conditional and the content that shows? Would also be much less error prone, you wouldn't be able to do the following:
```jsx
<li>
  <LinkExists to="jobs"> // blank <li> if jobs doesn't exist
     Jobs
  </LinkExists>
</li>
```
Could also re-use existing elements without lots of extra divs/spans

harrison:
Fixed: https://github.com/gocardless/splash-pages/pull/174

You always have to return an element (or null which turns into a noscript node) from a react component so attribute components won't work.

PR #174, opened 2015-05-29
Add tagName prop and return null from Translation

harrison:
You can now set `tagName` prop on `Message`, `Translation`, `InlineMessage`, `LinkExists`. Also changed `Translation` to return `noscript` node when it shouldn't be rendered.

### DIFF
--- a/app/components/footer/footer.js
+++ b/app/components/footer/footer.js
@@ -50,35 +50,25 @@ class Footer extends React.Component {
           <div className='grid__cell u-size-1of2'>
             <ul className='grid u-text-heading u-text-semi'>
               <div className='grid__cell u-size-1of3 u-text-xxs'>
-                <LinkExists to='features'>
-                  <li>
-                    <Link to='features' pointer='features.nav_title'
-                      id='track-footer-features' className='page-footer__link u-link-invert' />
-                  </li>
+                <LinkExists to='features' tagName='li'>
+                  <Link to='features' pointer='features.nav_title'
+                    id='track-footer-features' className='page-footer__link u-link-invert' />
                 </LinkExists>
-                <LinkExists to='pro'>
-                  <li>
-                    <Link to='pro' pointer='pro.nav_title'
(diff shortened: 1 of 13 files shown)


harrison:
@Ken-Lo what do you think?

@jamesshedden wanan review?

Ken-Lo:
@harrison This is epic!

Three quick thoughts:

1. This would be a massive breaking change, but worth it also be worth renaming `<LinkExists>` to `<IfLinkExists>`? Would enhance the readability somewhat. e.g.:
     `<IfLinkExists to='jobs' tagName='li'>`
2. Potential further enhancement to legibility (though not sure if it's actually possible to do in JS)? [JSX namespacing](https://github.com/facebook/react/pull/760) e.g. 

    ```jsx
  <IfLinkExists.li to='jobs'>
    Jobs
  </IfLinkExists.li>
    ```
Benefit is it keeps `li` in the name of the tag (more similar to what you expect when writing HTML). Downside is it's probably harder to implement (also not sure if Javascript supports overloaded functions)?
3. I think `<Message>` shouldn't be allowed to set a `tagName` or create new tags (it's only supposed to be used in an inline context anyway.)
e.g. this is confusing to read:

    ```jsx
<ul>
  <Message tagName='li' pointer='blah' />
  <Message tagName='li' pointer='blahblah' />
</ul>
```
or
    ```jsx
  <p><Message tagName='strong' pointer='phone_local' /></p>
```
(would be better to just specify this explicitly around the `<Message />` tag)

harrison:
Yeah `IfLinkExists` makes more sense, also removed `tagName` from message.
Namespaces could work with some funky getter to support arbitrary tagnames, but would want to play around with it before implementing.

Ken-Lo:
:+1: Nice one, loving this change!

Will leave it to you, @jamesshedden, to review :)

jamesshedden:
@harrison Awesome, :+1: from me!

PR #175, opened 2015-05-29
Update global expansion lead ad

jesssummerfield:
@jackfranklin please can you check this?

Fyi @Jennabrown adding in "After operationally setting this up..."


jackfranklin:
:+1:

PR #176, opened 2015-05-29
Show local phone number in footer

Ken-Lo:
Removes the international dialling prefix and shows a local number in the footer when available.

Germany footer is vastly improved:
screen shot 2015-05-29 at 12 26 36

France:
screen shot 2015-05-29 at 12 26 52

Europe/Ireland splash pages still shows number with international dialling prefix, they need to call us in the UK at present.
screen shot 2015-05-29 at 12 27 06

### DIFF
--- a/app/components/footer/footer.js
+++ b/app/components/footer/footer.js
@@ -163,7 +163,9 @@ class Footer extends React.Component {
                     <Message pointer='postal_address.address_country' />
                   </SchemaItemProp>.
                 </SchemaItemProp><br />
-                <SchemaItemProp itemProp='telephone' pointer='phone_full' />,&nbsp;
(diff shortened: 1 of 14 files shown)


Ken-Lo:
@anothertompetty Hey Tom,

As discussed earlier this morning, would be awesome to get your thoughts on this change. The IDD isn't required when calling a number within the same country (just confuses people). 

Possible disadvantage: People land on the wrong country version and then won't be able to contact us.

anothertompetty:
Looks :+1: to me.

Ken-Lo:
Nice one, thanks! Switching to local phone numbers across the whole website but adding the international phone number as an alternative on en-GB contact sales (this is the only locale with an appreciable number of people visiting the locale incorrectly).

PR #179, opened 2015-05-29
[DNM] One Contact Sales endpoint to rule them all

timrogers:
__DO NOT MERGE UNTIL https://github.com/gocardless/gocardless/pull/5991 IS MERGED__

At the moment, we have different `/contact-sales` endpoints for different countries - or more to the point, one for France and one for the UK. This unifies them and starts sending the locale and current path with the form, making everything magic.

At the moment also, the form is broken and the stuff that's meant to differentiate between French and UK leads doesn't work which is a bit rubbish. This fixes that too, when combined with changes in the GC app.

This means that the German contact sales page (and others for all the locales) will work as expected straight away. They just need buttons adding, which I'm not sure how to do.


harrison:
Neat! :+1:

jackfranklin:
:+1: nice!

PR #181, opened 2015-06-01
Add the partners list back to the FAQ

Ken-Lo:
Add the full list of partners back to the FAQ.


Ken-Lo:
Putting this up for discussion (I'm not sure what the background was for not moving this over to the new splash pages e.g. an accidental omission or a conscious decision)

@mchristofides Am I right in thinking you guys sometimes use this page?
@jamesshedden Did we make a decision not to copy this over to the new splash pages? Or did we choose to put it on the partners page instead (atm, we have a circular reference between the two pages)

jamesshedden:
@Ken-Lo Not sure why these were removed in favour of the circular link, good spot. Looks good apart from JSX needs `className=` instead of `class=` :+1:

Ken-Lo:
@jamesshedden Thanks - just wanted to double-check that we didn't mean to move it to /partners or something. Fixed the `className` attributes and merging...

PR #183, opened 2015-06-02
Clarified API FAQ for Irish and Europe merchants

octaveauger:
@Ken-Lo fyi, as discussed I've updated the API dev FAQ to take into account what is or is not currently available in the UK, Ireland and Europe - can you review the English to make sure it sounds native?


Ken-Lo:
@octaveauger Nice one! Looks good to me - I made some small changes and re-ordered the file so we could re-use common elements between each locale (rather than having a totally separate section for en-GB and for en-IE/en-EU).

Issue #168, opened 2015-05-26
Have a better way of expressing intent using the <Translation> tag


Ken-Lo:
So far, there are four main categories of intent when using the `<Translation>` component in the splash pages.

1. **Country-specific info**
Showing information that's specific to a certain country. For example: UK-only case studies or FR merchant logos. This intent is well served by the `<Translation>` component:
    ```jsx
    <Translation locales='en-GB'>
      UK specific info, logos, etc.
    </Translation>
    <Translation locales='en-IE'>
      IE specific info, logos, etc.
    </Translation>
    etc
    ```

2. **Domestic scheme & currency specific info**
Show different information depending on the domestic DD scheme or currency in that country.
    ```jsx
    <Translation locales={['en-GB']}> // Bacs
      Bacs, Direct Debit Guarantee, collect £s, etc.
    </Translation>
    <Translation locales={['en']} exclude={['en-GB']}> // SEPA
      SEPA, SEPA Customer Protection, collect €, etc.
    </Translation>
```
Examples: 
 * [app/pages/faq/merchants/direct-debit/faq-merchants-direct-debit.en.js:11](https://github.com/gocardless/splash-pages/blob/dev/app/pages/faq/merchants/direct-debit/faq-merchants-direct-debit.en.js#L11)
  * [app/pages/faq/merchants/security/faq-merchants-security.en.js:64](https://github.com/gocardless/splash-pages/blob/dev/app/pages/faq/merchants/security/faq-merchants-security.en.js#L64)

  This probably breaks down once we start adding additional schemes or countries (it becomes an ever-growing list of locales to specify in every file). 

3. **Signup process dependent info**
Show different information depending on whether self-serve sign-up is available (e.g. start taking payments now or contact sales CTA). At present, the intent is specified as follows:
    ```jsx
     // Here, {['en-GB']} is a white list of locales who can sign up immediately
    <Translation locales={['en-GB']}>
      Start taking payments today
    </Translation>
    <Translation locales={['en']} exclude={['en-GB']}>
      Contact sales
    </Translation>
```

  Examples:
  * [app/pages/faq/merchants/signing-up/faq-merchants-signing-up.en.js:13](https://github.com/gocardless/splash-pages/blob/dev/app/pages/faq/merchants/signing-up/faq-merchants-signing-up.en.js#L13)
  * [app/pages/features/features.en.js:120](https://github.com/gocardless/splash-pages/blob/dev/app/pages/features/features.en.js#L120)

[...]

PR #190, opened 2015-06-04
Add <IfLocale> component

Ken-Lo:
RE: https://github.com/gocardless/splash-pages/issues/168

This PR adds a `<IfLocale>` component. You can conditionally show information based on the local properties in each country (e.g. based on the domestic DD scheme or whether instant sign-up is available in the country).

```jsx
<IfLocale domesticScheme='bacs'>
  Stuff about Bacs Direct Debit
</IfLocale>
<IfLocale domesticScheme='sepa'>
  Stuff about SEPA Direct Debit
</IfLocale>
<IfLocale domesticScheme='bankgirot'>
  Stuff about Bg Autogiro
</IfLocale>

<IfLocale hasInstantSignup>
  Signup now in 5 minutes
</IfLocale>
<IfLocale hasInstantSignup={false}>
  Contact Sales
</IfLocale>

<IfLocale hasPercentagePricing>
  % pricing is available
</IfLocale>
<IfLocale hasPercentagePricing={false}>
  % pricing is not available
</IfLocale>
```

In order to make it work, the `country_properties` message needs to be set for each locale in the app/messages/ folder. e.g.

```js
  country: 'United Kingdom',
  country_properties: {
    local_scheme: 'bacs',
    has_instant_signup: true,
    has_percentage_pricing: true,
  },
```

```js
  country: 'Belgique',
  country_properties: {
    local_scheme: 'sepa',
    has_instant_signup: false,
    has_percentage_pricing: true,
  },
```


Ken-Lo:
@harrison Would be good to get your thoughts on this approach.

Also thinking about possibly generalising this component to `<IfLocal>` and allowing it to work on other country-based properties (e.g. whether immediate sign-up is available, whether Basic pricing is available)

app/components/if-local-scheme/if-local-scheme.js
```diff
@@ -0,0 +1,39 @@
...
+function getLocalesWithLocalScheme(selectedScheme, availableLocales, countryProperties) {
+  var localesWithScheme = [];
+
+  for (var locale in countryProperties) {
+    if (countryProperties[locale].local_scheme === selectedScheme) {
+      localesWithScheme.push(locale);
+    }
+  }
+
+  return localesWithScheme;
```
jackfranklin: you could use `filter` from LoDash here:

```js
return _.filter(countryProperties, function(locale) {
  return locale.local_scheme === selectedScheme;
});
```

jackfranklin:
@Ken-Lo welcome to the frontend team! ;) This looks really awesome. I'll let @harrison review other than the one thing I spotted (happy to chat through further) but just wanted to say that this seems like a really nice, intuitive improvement that will help a lot :+1:

app/components/if-local-scheme/if-local-scheme.js
harrison: :+1: 
but think you will need to return the object key:
```
return Object.keys(countryProperties).filter(function(locale) {
  return countryProperties[locale].local_scheme === selectedScheme;
});
```

app/components/if-local-scheme/if-local-scheme.js
```diff
@@ -0,0 +1,39 @@
...
+    children: React.PropTypes.node.isRequired,
+  }
+
+  static contextTypes = {
+    availableLocales: React.PropTypes.array.isRequired,
+    countryProperties: React.PropTypes.object.isRequired,
+  }
+
+  render() {
+    const { availableLocales, countryProperties } = this.context;
```
harrison: `availableLocales` is unused

app/components/if-local-scheme/if-local-scheme.js
harrison: Wait do you need this?

Could you do this in render?

```js
const schemes = getMessage(messages, 'country_properties');

 if (schemes. local_scheme !== this.props.name) { return null; }
return React.createElement('span', this.props, this.props.children);
```

harrison: think you could then remove `countryProperties`

Ken-Lo: Nice one @jackfranklin, I knew there must have been a better way of selecting items from an array (didn't find anything in the MDN docs). Thanks for the info about lodash!

@harrison Makes sense, if we aren't wrapping `<Translation>`, then we wouldn't need to know the properties of other countries. Will move to render and send you a v2 :)

Ken-Lo:
@harrison Just updated this:

* Moved everything to render() so no need to pass countryProperties in the context
* Generalised to `<IfLocale>` so we can also do this on other country properties
* Added some tests
* Updated a couple of new pages to use the `<IfLocale>` tag

Would be good to get your thoughts!

app/components/if-locale/if-locale.spec.js
```diff
@@ -0,0 +1,163 @@
...
+          <TestIfLocale hasPercentagePricing>
+            foo
+          </TestIfLocale>
+        );
+
+        expect(renderText(component)).toEqual(null);
+      });
+    });
+  });
+
```
harrison: :scissors: empty lines

app/components/if-locale/if-locale.spec.js
```diff
@@ -0,0 +1,163 @@
...
+  if (elements.length === 0) {
+    return null;
+  } else if (elements.length === 1) {
+    const element = ReactTestUtils.findRenderedDOMComponentWithTag(result, 'span');
+    return element.getDOMNode().textContent;
+  }
+}
+
+describe('IfLocale Component', () => {
+
```
harrison: :scissors: empty lines

harrison:
Nice work! :+1: minor comments

harrison:
Closes #168

Ken-Lo:
@harrison Thanks! Merging this now, will tidy up the FAQ & other pages to use `<IfLocale>` in a separate pull request.

PR #202, opened 2015-06-09
Fix redirects written after glob

harrison:


harrison:
Writing the redirect files and the glob is all sync but the glob doesn't pick up the redirect file.. and running it a second time works fine:(

Not sure why this is the case... @jackfranklin run into anything similar?

PR #205, opened 2015-06-11
Static site generation optimisation

domderen:
While preparing to exclude your this static page generation i thought of this small optimisation, and I thought you guys might be interested in it.

I am sending this PR mainly for the purposes of the peer review to see if I didn't miss any reasons why you used express in your static page generation process,
and to see if it could indeed be skipped, like I did here.

Thanks in advance for any feedback.

Static page rendering optimisation
- Bypassing express server in static site generation,
- dump-html now uses the same render function, as the express middleware,
- removed fetchPage method, as it was now not required,
- simplified build-prod script to only run production web pack build, and html-dump, no express sever in the middle.

### DIFF
--- a/app/static-build/builder.spec.js
+++ b/app/static-build/builder.spec.js
@@ -1,39 +1,13 @@
-import nock from 'nock';
 import fs from 'fs';
 import rimraf from 'rimraf';
 import mkdirPCallback from 'mkdirp';
-import { fetchPage, writePage } from './builder';
+import { writePage } from './builder';
 import denodeify from 'denodeify';
 
 const readdir = denodeify(fs.readdir);
 const readFile = denodeify(fs.readFile);
 const rmdir = denodeify(rimraf);
 const mkdirP = denodeify(mkdirPCallback);
-
-describe('fetchPage', () => {
-  const host = 'http://localhost:1234';
-
-  beforeEach(() => {
-    nock(host)
-    .get('/')
-    .reply(200, '<html>home</html>')
-    .get('/about')
-    .reply(200, '<html>about</html>');
-  });
-
-  afterEach(nock.cleanAll.bind(nock));
-
-  it('resolves [url, body] tuple', (done) => {
-    const path = '/about';
-
-    fetchPage(host, path)
-      .then(function([url, body]) {
-        expect(url).toEqual(`${host}/about`);
-        expect(body).toEqual('<html>about</html>');
-      })
-      .then(done, done.fail);
-  });
-});
 
 describe('writePage', () => {
   beforeEach((done) => {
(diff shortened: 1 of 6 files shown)


harrison:
Nice idea! Haven't had a chance to properly look this over yet, will do it in the next couple of days.

server/render.js
```diff
@@ -27,69 +27,90 @@ function normalizeUrl(urlStr) {
...
+        <HtmlDocument
+          markup={markup}
+          script={webpackUrls.script}
+          css={webpackUrls.css}
+          router={router}
+          dataRender={appProps(stateProps)}
+          {...appProps(stateProps)} />
+      );
+      const doctype = '<!DOCTYPE html>';
+      resolve([requestUrl, doctype + html]);
```
harrison: think we can just do `resolve(doctype + html)` - this format is used by `writeHtml` but would be better to deal with it in `dump-html`

server/render.js
```diff
@@ -27,69 +27,90 @@ function normalizeUrl(urlStr) {
...
-        script={webpackUrls.script}
-        css={webpackUrls.css}
-        router={router}
-        dataRender={appProps(stateProps)}
-        {...appProps(stateProps)} />
-    );
-    const doctype = '<!DOCTYPE html>';
-    res.send(doctype + html);
+  render(req.url)
+  .then(([, htmlOutput]) => {
```
harrison: `.then((htmlOutput) => {`

script/dump-html
```diff
@@ -1,23 +1,24 @@
...
-  console.error(`Usage: ${bin} ${script} host directory`);
+if (!outDir) {
+  console.error(`Usage: ${bin} ${script} directory`);
   process.exit(1);
 }
 
 const processPage = function(pagePath) {
-  return fetchPage(serverUrl, pagePath)
+  return render(pagePath)
     .then(writePage.bind(null, outDir))
```
harrison: `.then((htmlContent) => { return writePage(outDir, [pagePath, htmlContent]); })`

app/static-build/builder.spec.js
```diff
@@ -1,40 +1,14 @@
-import nock from 'nock';
```
harrison: could remove `nock` from the package.json

script/build-prod
```diff
@@ -1,35 +1,18 @@
...
-
-./script/dump-html "http://localhost:$PORT" 'dist'
+./script/dump-html 'dist'
 DUMP_EXIT=$?
 
-kill $SERVER_PID
-
 set -e errexit
 
 if [ $DUMP_EXIT -eq 0 ]; then
```
harrison: think this needs to be resolved in `script/build` now

megaslug:
Closing to address @harrison comments [here](https://github.com/gocardless/splash-pages/pull/837)

PR #210, opened 2015-06-11
Update German splash pages to link to German Guides

Ken-Lo:


Ken-Lo:
FYI @jutta-frieden 
Will merge this once we've deployed the German guides into production (need to keep the repo deployment in sync)

PR #215, opened 2015-06-18
made German CL job advert more salesy

jutta-frieden:
Made it more focused on the sales part. Also, I mentioned that this will be a role together with a second Country Lead Germany.


jutta-frieden:
@jesssummerfield could you proof-read and give your ok before we merge this? Thanks!

PR #223, opened 2015-06-25
added Basic pricing translations

jutta-frieden:


jutta-frieden:
@Ken-Lo could you check everything is ok from your end? Thanks!

Ken-Lo:
@jutta-frieden Looks good to me! One more thing: we might need to update the "Gibt es wirklich keine weiteren Gebühren?" section (it currently mentions only the Pro pricing). If you like, we can use something like the following (taken from [here](https://gocardless.com/en-ie/pricing/)):

> **Are there any other fees?**
No, there are no other fees except from those listed in the table above. We never charge for chargebacks or payment failures.
With GoCardless, all you ever pay is 1% of the transaction amount, up to a maximum of €2.
With GoCardless Pro, we charge you the €250 monthly fee and a fee for each transaction.

jutta-frieden:
@Ken-Lo Done. note that I changed the way we display euro amounts in the messages: e.g. .from €2 to 2€ and from €250 to 250€.

Ken-Lo:
:+1: Looks good, thanks @jutta-frieden!

PR #224, opened 2015-06-26
update sandbox-signup link

samjewell:
Point sandbox signup to `manage-sandbox.gocardless.com/signup`


harrison:
Nice :+1: is the signup tested for sandbox or is it all the same flow as prod?

samjewell:
@harrison Not sure I understand - do you mean the end-to-end integration tests that we run, and get emails about if/when they fail?

harrison:
Ignore, was thinking if we did any special casing for signups in prod/sandbox but seems like it's just the same flow!

harrison:
on the backend

PR #225, opened 2015-06-29
Use CSS columns for links in footer

Ken-Lo:
I don't necessarily want to merge this, though would be good to get some feedback on whether it's a good idea.

Currently, we hard-code links in the footer into three separate columns. This can sometimes give you slightly odd-looking layouts, depending on the pages available in the locale.

**France:** (5/3/4, would look better as 4/4/4)
screen shot 2015-06-29 at 14 19 54

**Ireland:** (5/3/4, would look better as 4/4/4)
screen shot 2015-06-29 at 14 20 25

**Germany:** (5/1/3, would look better as 3/3/3)
screen shot 2015-06-29 at 14 20 33

One way of getting around this is to use CSS columns rather than grids.

**France with CSS columns:**
screen shot 2015-06-29 at 14 24 34

**Ireland with CSS columns:**
screen shot 2015-06-29 at 14 24 46

**Germany with CSS columns:**
screen shot 2015-06-29 at 14 25 12

The disadvantage of doing this is it doesn't work in around 6% of browsers. Specifically, [it doesn't work in IE9 or earlier](http://caniuse.com/#feat=multicolumn)).

One other way of getting around this (more complex but would probably also work on IE9 & earlier) is to populate the three grid cells using Javascript instead.

Any thoughts?


Ken-Lo:
@harrison @jamesshedden Just wondering if you guys had any thoughts on the above? Should be slightly cleaner & more maintainable as we go to new countries but could also break in some older web browsers that don't support the feature (IE9 or earlier)

harrison:
yeah would be cool with this solution, makes it better for like ~97% (looking at ga) of our users and not crazy worse for users on old ie

Ken-Lo:
@harrison Cool! How would you go about implementing this? Something like `.u-columns-2`, `.u-columns-3` in app/css/utils/ or just a special case in app/css/components/footer.scss?

jamesshedden:
@Ken-Lo Nice idea! Re: implementation, generally utility classes are created out of necessity (ie there's a demand for an app-wide useful class that can be reused all over the place) so for time being I'd stick to classes that are specific to the footer.

Ken-Lo:
@jamesshedden Thanks! I've moved the styles to a `.footer__links` CSS class. Have also added a 20px column gap to match the styling on the old `.grid__cell`. Any other changes or OK to merge?

The only other place where I think we currently use CSS columns is [here in the SEPA Guide](https://gocardless.com/guides/sepa/what-is-sepa/), so in a totally different repo. Will update the Guides in a separate PR.

jamesshedden:
@Ken-Lo Nice :+1:

Ken-Lo:
@jamesshedden Thanks for the feedback! As discussed at https://github.com/gocardless/guides/pull/223, have deleted the now redundant footer.scss file and have renamed the class to `.page-footer__link-list` to promote cross-repo consistency.
Let me know if you have any thoughts!

jamesshedden:
@Ken-Lo Awesome work :+1:

PR #226, opened 2015-07-01
Adding support ad

jesssummerfield:
@jackfranklin please can you check this?

fyi @bamesjerry @erezmt

### DIFF
--- a/app/messages/en.js
+++ b/app/messages/en.js
@@ -394,6 +394,10 @@ export default {
     title: 'Hiring Coordinator',
     description: '',
   },
+  jobs_customer_support: {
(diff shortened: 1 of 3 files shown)


app/messages/en.js
```diff
@@ -394,6 +394,10 @@ export default {
     title: 'Hiring Coordinator',
     description: '',
   },
+  jobs_customer_support: {
+    title: 'Customer Support',
+    description: '',
+  },
```
jackfranklin: I think you might need to add a similar message to `app/messages/fr.js`, as this job exists in both FR and EN locales.

PR #230, opened 2015-07-03
Never hardcode a sign up link, and always use messages.

jackfranklin:
Keeps us nimble should stuff change.

@samjewell can I get your eagle eye on this please?


samjewell:
:+1: good work - just failing some tests :-(

Ken-Lo:
@jackfranklin This is awesome! Would it be good practice or overkill to fail tests when someone links to directly to signup? Would actually be interested in seeing how feasible this is, it could come in super handy for phone numbers, etc (we might want to later vary the signup link or phone number based on acquisition channel/campaign)

jackfranklin:
@Ken-Lo yeah, I'd like to look into that. One for the backlog!

PR #244, opened 2015-07-15
Fix irish login

petehamilton:
@jacobpgn or @isaacseymour review?

### DIFF
--- a/app/messages/en-GB.js
+++ b/app/messages/en-GB.js
@@ -9,6 +9,12 @@ export default {
   },
   phone_local: '020 7183 8674',
   documentation_link: 'https://developer.gocardless.com/',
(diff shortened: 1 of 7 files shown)


jackfranklin:
We should investigate how we can have some form of tests for these. Tests should ensure we don't manage to push anything where there isn't a path set that's mentioned in the HTML. I would have thought this would have broken when running locally. Will look into it tomorrow.

:+1: though

isaacseymour:
There were a few other places using `/merchants/new` hard-coded. I've replaced them all with `Href to='signup.path'`, so this shouldn't be an issue anywhere else either :)

jackfranklin:
:+1:

PR #245, opened 2015-07-16
[Do Not Merge] Output an XML sitemap from the splash pages

Ken-Lo:
Please do not merge - this is a work in progress (would be good to get some thoughts/comments regarding the approach).

**Approach:**

1. Loop through all the JS files in `app/pages/` and `/app/components/`:
  * The last modification time for the file is obtained by running `git log --format='%ct' -- [filename]`. The script we use to generate XML sitemaps on the guides [uses this approach](https://github.com/gjtorikian/jekyll-last-modified-at/blob/master/lib/jekyll-last-modified-at/determinator.rb#L28) as filesystem mtime is not accurate.

    ```js
    var moduleModificationTimes = {
      '/Users/khlo/splash-pages/app/pages/stories/stories.js': 1432343039
    }
    ```
  * Store the name of each module in a lookup table so we can find the file it's contained within e.g.

    ```js
    var moduleNames = {
      'Stories' : '/Users/khlo/splash-pages/app/pages/stories/stories.js',
    }
    ```

2. Get a list of all possible routes from `app/router/routes.js`.
  * For each route, find the main JS file corresponding to that route. e.g.

    ```js
    var routeName = 'Stories';
    var filename = getFilenameForRoute(routeName);
    // Uses the moduleNames lookup table to give /Users/khlo/splash-pages/app/pages/stories/stories.js
    ```

  * Resolve a list of all import dependencies (imported components & pages).

    ```js
    // Will resolve a list of all dependencies for /Users/khlo/splash-pages/app/pages/stories/stories.js
    var allDependenciesForRoute = buildDependencyTree(filename);
   ```

  * Find the mostly updated dependency for each route.

    ```js
    var fileModificationTimes = allDependenciesForRoute.map((dependency) => moduleModificationTimes[dependency]);
    var routeLastModificationTime = Math.max(...fileModificationTimes);
    ```

   * The modification time of the most recently updated dependency = the last modified time for that page.

**Notes:**

To generate sitemap,  `npm run build-xml-sitemap`.


Ken-Lo:
Hey @english 

As discussed earlier this morning... would be awesome to get your thoughts on doing something like this. Would probably need a bit of re-factoring, tidy-up etc.

Ideally, would be awesome to get something like this put into the static site build process (so it dynamically generates a sitemap.xml file every time we deploy the splash pages).

script/build-xml-sitemap
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * Returns the last modification time of a file according to the latest commit in git log.
+ * Runs `git log --format='%ct' -- [filename]` and returns the first line
+ * @param {string} filename e.g. app/pages/about/about.js
+ * @return {integer} e.g. 1432343039
+ */
+function getFileLastModificationTime(filename) {	
+  // Could modify this to use fs.statSync(file) & read mtime from file system if we aren't using git
+	var lines = execSync("git log --format='%ct' -- "+filename).toString().split("\n");
```
Ken-Lo: Downside: Running `git log --format='%ct'` on every single file isn't the fastest process (takes ~1 min to generate the sitemap on my local computer). Could probably speed things up by making this async or something.

english:
@jackfranklin I'd be interested to get your thoughts on this and any alternative approaches.

script/build-xml-sitemap
jackfranklin: You can use `fs.stat` (https://nodejs.org/api/fs.html#fs_fs_stat_path_callback) to do this, quicker than going through Git. One of the things it gives you back is the file's `mtime`.

jackfranklin: Sorry, just read your comment which said exactly what I said in the above comment, sorry!

jackfranklin:
@english @Ken-Lo will take a look when I'm back on Tues :)

script/build-xml-sitemap
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * Get a list of all Javascript imports in a file (uses ES6_IMPORT_REGEX for matching import statements).
+ * Based on https://gist.github.com/pilwon/ff55634a29bb4456e0dd. Modified the regexp to extract path of file.
+ * Will only return files from the allModule global variable (contains a whitelist of permitted files).
+ * @param {string} filename e.g. app/pages/stories/stories.js
+ * @return {array} e.g. [ '/Users/khlo/react-splash-pages/app/pages/stories/story-link.js' ]
+ */
+const ES6_IMPORT_REGEX = /\bimport\s+.+\s+from\s+[\'"]([^"\']+)["\']/;
+function getListOfJavascriptImports(filename) {
```
jackfranklin: first thought that comes to mind would be that it would be kind of fun to use an AST to find these rather than a regex. Could maybe be a tiny bit overkill though at the same time.

english:
Nice work @Ken-Lo! The code looks great, however I'm not sure of the high-level approach.

1. We'll miss changes that cut across the whole site, like header and footer changes (am I correct here?).
2. Relying on 'pages' always living in `app/components` or `app/components` is potentially brittle.
3. Building a dependency tree using regular expressions for `import` statements feels a bit brittle too. This is less of an issue than the above two.

I've been thinking about this over the weekend. What would you think about committing our built html files to `git` (either in this repo or a separate one) and then building the sitemap as a build step? While it might sound a bit crude, it would be 100% accurate for a) what pages are there and b) their last modification date (using your `git log` method). It would also be entirely decoupled from our javascript implementation of pages (which `js` files lead to which pages, and where they live in the file system), no need to do anything complicated around dependency trees.

@Ken-Lo, @jackfranklin what do you guys think?

script/build-xml-sitemap
Ken-Lo: Hey @jackfranklin 
I don't think the file system `mtime` is accurate - git will change it every time it modifies a file ([1](https://git.wiki.kernel.org/index.php/GitFaq#Why_isn.27t_Git_preserving_modification_time_on_files.3F), [2](http://stackoverflow.com/questions/1964470/whats-the-equivalent-of-use-commit-times-for-git/1964508#1964508)). So it only tells you when the file was changed on your local computer (including from git checkouts, changing branches, etc) and doesn't tell you when it was changed in the repo.

Ken-Lo:
Hey @english 

Thanks for the feedback! Happy to do it either way depending on what you guys prefer (though would obviously be a bit of a shame to have all the overhead of a new repo just for generating the sitemaps). 

As you say, the benefit of doing so is it would be much more reliable (e.g. re-factoring of the JS which doesn't affect the final HTML output wouldn't cause it to give a totally new modification time). 

A possible complication: `window.app` and the JS bundle appear on every page ( e.g. `<script>window.app={"currentLocale":"en-GB","language":"en","messages":{"header":{"our_products":"Our products"...}}}</script><script src="/bundle/main-6a631a84a9e42b24aafc.js"></script>`). Presumably this means the HTML page will always change upon site re-generation (it certainly changes if window.app changes, not sure how the bundle file name also changes?)

In response to your points:

1. It does currently pick up changes to the header & footer (they get pulled in as a dependency via `app/components/page/page.js`).
2. This is true, though we could modify the glob statement to consider everything within the repository (I think it would give the exact same result as we don't actually import anything from outside these folders?)
3. Agree this is quite brittle. This line is also quite brittle: `var moduleName = require(filename).name;` as it assumes a module is exported from that file. Could probably use an AST like Esprima to make import detection & module name declaration more rigid?

english:
@Ken-Lo thanks for addressing those points.

Good point about the changing asset filenames - I hadn't thought of that! I suppose a solution would be to strip out `script`s in the generated output before committing. That feels a bit hacky and could be enough to rule out the separate git repo option.

Ken-Lo:
^ I've just updated the code to run `git log --format='%ct' -1` (only returns the timestamp of the most recent commit). Also using `Promise.all()` which speeds it up to finish in ~5 seconds (thanks Jamie!).

Will leave the rest of it on hold so you guys can have a think about dependency trees, etc

jackfranklin:
These may be stupid questions as I've been out the loop of discussions, please feel free to correct me...

1. If we do stick to the current approach, we should definitely use an AST to pull out the import statements rather than regexes. This could even make a nice standalone thing. The tool could take a file, crawl all imports and produce a tree of dependencies for that given file.

2. Is it possible to use React to generate XML? It seems crazy to be generating all this stuff ourselves; if we have a bunch of React components already, can we harness React some more to generate the sitemap?

3. If the generated HTML files were to include the latest modified time in the output (`body data-mtime="10202020343948"`), we could run a small script after the build process that uses these? We could move the logic that figures the last modified time into the application and then write one that can take the outputted HTML files and generate the sitemap from those? Is that even more complex though?

4. I'm against keeping the HTML files in git quite strongly if we can avoid it. It doesn't make sense in my mind for the HTML files to be kept in git; given they are generated from the contents of this repo and should be pure (ie, the same git repo generates the same HTML, every time), I'd like to not keep the HTML in the repo. It also then becomes easy for the HTML to get out of sync.

What are your thoughts?

Ken-Lo:
Hey Jack - thanks!

1. This does sound like it would be helpful as a standalone project!

2 & 3. This is a really awesome idea but sounds like it could be much more complicated! It would, however, solve a problem we have with the existing setup (the last modification time can be updated by an included component that doesn't actually render anything).

Example: the `<Pricing />` component currently includes `<PricingEn />`, `<PricingFr />`, etc. When accessing `/fr/tarifs`, the `<PricingEn />` component renders nothing (everything comes from `<PricingFr />`). With the current setup, an update to `pricing.en.js` would bump the last modification time for pricing pages in every country (even those where `<PricingEn />` doesn't render a thing).

One way of solving this problem would be to add a `data-mtime` attribute to all rendered components. I guess this would probably be done by extending [the current code which generates data-reactid on the server](https://github.com/facebook/react/blob/59dfe87d0419d1401aed82631e49dac34011a89b/src/renderers/dom/server/ReactServerRendering.js#L48). You could then extract all the `data-mtime` attributes and potentially run a Math.max() on every file in the static site build. I have no idea if this is actually possible without modifying React itself?

jackfranklin:
It also just occurred to me that the ability to take a file and trace its imports has been written already in SystemJS builder - in particular its trace API: https://github.com/systemjs/builder#example---direct-trace-api. If it exposes its tree that it gets, we could use that instead? Worth a look.

We would have to modify React which is something we definitely don't want to do. It sounds like doing what you've done already but swapping out the regexes to properly trace the tree would be the right direction.

english:
@Ken-Lo I had a chat with @lawrencejones and @jackfranklin earlier and we've got an idea that we think makes sense, I'd be interested to hear what you think:

1. We build the static HTML and strip out our `<script>` and `<link>` tags
2. We fetch the currently deployed sitemap (at gocardless.com/sitemap.xml or whatever)
3. As we're building the new sitemap (by crawling the static HTML files), we diff fetch the currently deployed static file (stripping `<script>`s and `<link>`s) and diff it with our newly built one.
4. If our new HTML is different to the deployed one, our `lastmod` is today, if its the same we use the `lastmod` from the deployed sitemap.

The advantages of this approach are:
- We keep no 'non-source' files in git.
- Its entirely decoupled from the implementation (doesn't have to be a react app, or even JS).

Disadvantages:
- Stripping `<script>` and `<link>` tags is a bit hacky

It would be really nice if we could add custom 'metadata' to the deployed `sitemap.xml`. If we could, it would mean we could hash the contents of the static files (without fingerprinting assets) and add that to the sitemap. We could then hash our new html and compare the hash with the one in the sitemap, thus avoiding stripping tags.

I've had a quick look on [sitemap schema](http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd) and it doesn't seem to allow extra tags or attributes. @Ken-Lo do you know if this is definitely the case? Would it be worth testing?

jackfranklin:
> Stripping <script> and <link> tags is a bit hacky

It's really easy to do actually with something like Cheerio or similar, so it can be done in a way that's not so hacky.

Ken-Lo:
@english Nice - sounds like a workable & neat solution! Several quick thoughts:

1. It introduces hysteresis to the build process (it's no longer possible to build our set of output files simply by knowing the input files for that build: you also need to know the result of your previous build). This isn't necessary a problem but is something worth thinking about (especially if this behaviour will be unique amongst all of our repos).

2. It just occurred to me that we'd probably also need to strip `data-reactid` attributes (they appear to be randomly generated during the build). Might be easier or neater to simply strip out all of the HTML before generating the fingerprint - thereby you'd only compare the text on the page as opposed to other HTML within that file. Web crawlers only really care about the text in a document anyway so this allows us to diff only the most important parts.

3. I suggest storing the hash values in a separate file (perhaps local to the sitemap generation process). That way, you can entirely decouple the sitemap generation process from being able to write to that server or sitemap. If you do this, there'd be no actual reason why you couldn't use the script to generate sitemaps for any website in the world (e.g. you could generate a sitemap for bbc.co.uk with `<lastmod>` times, only you would need different logic to get the list of URLs).

4. If we choose to store the hash values in the sitemap itself, it shouldn't be a major problem... The sitemap schema doesn't allow custom tags but you can always use a custom XML namespace. Google does this itself by [importing the XHTML schema for international sitemaps](https://support.google.com/webmasters/answer/2620865?hl=en). As a worse case scenario, you can also store the hash value in a XML comment (still accessible through [a XPath selector](http://stackoverflow.com/questions/784745/accessing-comments-in-xml-using-xpath)).

lawrencejones:
Producing a last modified value on our HTML requires some concept of what that actually means. From the above and conversations with @english, I've taken it to mean that a modification has occurred to the page that is apparent in it's textual content.

If this is the case, then parsing the HTML output with something like cheerio, we could find a hash of `$('body').text().replace(/\s/g, '')` and use that to represent the state of the page for the purpose of a last modified calculation.

Unfortunately calculating whether something has changed requires us to have knowledge of what came before it's current state. Carrying over last modified values is the only way of doing this without having to run the computation across the repo history every time we generate the sitemap, which would be prohibitively expensive (and not possible, assuming the calculation is coupled to our implementation of splash pages).

Instead of generating this at build time, I suggest we add a git hook that warns people before committing that their sitemap is out of date. They can then choose to re-run the generation, though we can enforce at a circle level that the sitemap is consistent with the current site, ensuring nothing hits production that lacks an up-to-date sitemap. This ensures we don't let the last modified values grow steadily more out of date, and blocking builds if a repo artefact isn't consistent with HEAD is something we do on PS already with JSON schemas.

Assuming that we have this circle check enforced, we can then use git to fetch the checked in sitemap from our previous commit rather than relying on what's already published. This allows us to do the calculation all inside our repo which I favour over having to depend on something hosted online, even if it is by us.

^^ Just some thoughts that I've had since @english mentioned this yesterday.

Ken-Lo:
@lawrencejones Great analysis! And your suggestion for how to determine when the content has changed sounds great to me.

Quick q: How would the git hook work in practice? I presume when you merge dev or master, it does a static site build for both the before and after versions and then diffs the two to find out which pages have changed? If we can automate this to run in Circle and if it doesn't affect the normal git workflow, it sounds like a good option to me.

It would be good to get a sense of how much dev work the options above would all take? If it's a really big job, we obviously shouldn't allow this to distract from Enterprise Dashboards (while sitemaps are something we should have, it's not so transformative that we should spend major dev effort making it all work).

lawrencejones:
@Ken-Lo talked with @english and we're not sure what the ideal solution is. I actually think that using the git last modified search would probably do fine for our use case after giving it some real thought about if false positive cost us too much. @english will have more of an idea however.

And for the git hooks, I mean probably use a `pre-commit` hook to run the check before people go ahead to commit. This means they'd get some warning that what they are about to check in won't pass a circle test, while allowing them to commit anyway if they wanted to skip building the sitemap (if you wanted to push a WIP before heading home for example). The circle check would then be enforced on every build to verify that the sitemap is up to date.

english:
@Ken-Lo come and grab me when you're next in and we can have a chat about it. It looks like nobody is 100% happy with any solution, so we should weigh up the benefit vs maintenance cost of each of our available approaches.

jackfranklin:
@english I'd like to be in on that too please - agree chatting in person is going to be more productive than GitHub from this point!

english:
Yeah, absolutely @jackfranklin. @lawrencejones if you fancy joining too, just shout!

Ken-Lo:
@english @jackfranklin @lawrencejones Nice one - cheers guys! Let's have a quick catch-up tomorrow morning... I've put in a bit of time at 10.30 to discuss!

lawrencejones:
I'm afraid I probably won't be able to make that tomorrow, got a meat order arriving at my place in the morning so will be a late start for me :( I may join halfway through, depends on when my food arrives!

Ken-Lo:
Superseded by https://github.com/gocardless/splash-pages/pull/251.

PR #251, opened 2015-07-27
Output an XML Sitemap From Splash Pages - v2

Ken-Lo:
A second version of our XML sitemap generation following the discussion last week.

Creates a XML sitemap file from the list of all routes as part of the static site build.
* Calculates a MD5 hash of the page by stripping the Javascript and running `$('body').text()`
* Loads in the currently deployed version of the sitemaps (gocardless.com/sitemap.xml for master, staging.gocardless.com/sitemap.xml for dev).
  * Where the MD5 hash for a page hasn't changed since the previous deploy, the `<lastmod>` value is kept from the old sitemap.
  * Where the MD5 hash value differs, use the current date for `<lastmod>`

### DIFF
--- a/app/static-build/sitemap.js
+++ b/app/static-build/sitemap.js
@@ -1,0 +1,150 @@
+import url from 'url';
+const cheerio = require('cheerio');
+const crypto = require('crypto');
+
+/**
+ * Creates a new sitemapUrlSet object.
+ * e.g. var sitemap = new sitemapUrlSet('http://example.com')
+ *
+ * @class
+ * @param {string} domain - Domain to which the sitemap refers e.g. https://example.com
+ */
+export class sitemapUrlSet {
+  constructor(domain) {
+    this.urlSet = {};
+    this.domain = domain;
+  }
+
+  /**
+   * Loads in an existing XML Sitemap and returns a sitemapUrlSet object.
+   * e.g. var currentSitemap = new sitemapUrlSet('http://example.com').fromXml(xmlFileContents)
+   *
+   * @param {string} xml - Sitemap in XML format
+   * @returns {sitemapUrlSet} A sitemapUrlSet object representing the sitemap.
+   */
+  fromXml(xml) {
+    var sitemapUrlSet = this;
+    var $ = cheerio.load(xml);
+    var sitemapUrls = $('url');
+    sitemapUrls.each(function(i, element) {
+      var location = $(this).find('loc').text();
+      if (location) {
+        var pathName = url.parse(location).pathname;
+        var hash = $(this).find('content\\:hash').text();
+        var lastMod = Date.parse($(this).find('lastmod').text());
+        sitemapUrlSet.urlSet[pathName] = {
+          'hash': hash,
+          'lastMod': lastMod,
+        };
+      }
+    });
+    return sitemapUrlSet;
+  }
+
+  /**
+   * Returns an XML representation of the current sitemapUrlSet object.
+   *
+   * @returns {string} A XML format representation of the sitemapUrlSet object.
+   */
+  toXml() {
+    var xml = "<?xml version='1.0' encoding='UTF-8'?>\n"
+            + "<urlset xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'\n"
+            + "  xsi:schemaLocation='http://www.sitemaps.org/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd'\n"
+            + "  xmlns='http://www.sitemaps.org/schemas/sitemap/0.9'\n"
+            + "  xmlns:xhtml='http://www.w3.org/1999/xhtml'\n"
+            + "  xmlns:content='https://github.com/gocardless/splash-pages#xml-sitemap-hash'>\n";
+    for (url in this.urlSet) {
+      var lastMod = new Date(this.urlSet[url]['lastMod']);
+      var lastMod_w3 = lastMod.toISOString().substring(0,10); // Date should be in W3C Datetime (YYYY-MM-DD) format
+      xml += `  <url>\n`
+          +  `    <loc>${this.domain}${url}</loc>\n`
+          +  `    <content:hash>${this.urlSet[url]['hash']}</content:hash>\n`
+          +  `    <lastmod>${lastMod_w3}</lastmod>\n`
+          +  `  </url>\n`;
+    }
+    xml += '</urlset>'
+    return xml;
+  }
+
+  /**
+   * Add a URL to the sitemap.
+   * e.g. sitemap.addUrl('http://www.example.com/about/', 'About Page Contents')
+   *
+   * @param {string} pageUrl - URL of the page to add to the sitemap
+   * @param {string} content - HTML contents of the page
+   */
+  addUrl(pageUrl, content) {
+    const pathName = url.parse(pageUrl).pathname;
+    const pageHash = getHashOfContent(content);
+    this.urlSet[pathName] = {
+      'hash': pageHash,
+      'lastMod': Date.now(),
+    }
+  }
+
+  /**
+   * Extracts a URL from the sitemap.
+   * e.g. var aboutPage = sitemap.getUrl('http://www.example.com/about/')
+   *
+   * @param {string} pageUrl - URL of the page to add to the sitemap
+   * @return {{Object|boolean}} Returns URL info if exists in the sitemap; otherwise returns false 
+   */
+  getUrl(pageUrl) {
+    const pathName = url.parse(pageUrl).pathname;
+    if (pathName in this.urlSet) {
+      return this.urlSet[pathName];
+    }
+    return false;
+  }
+
+  /**
+   * Removes a URL from the sitemap.
+   * e.g. sitemap.removeUrl('http://www.example.com/about/')
+   *
+   * @param {string} pageUrl - URL of the page to remove from the sitemap
+   * @return {boolean} Returns true if page was removed from sitemap; otherwise returns false
+   */
+  removeUrl(pageUrl) {
+    const pathName = url.parse(pageUrl).pathname;
+    if (pathName in this.urlSet) {
+      delete this.urlSet[pathName];
+      return true
+    }
+    return false;
+  }
+
+  /**
+   * Imports the timestamps from an old sitemap when the content is unchanged.
+   * e.g. newSitemap.importTimestampsFromOldSitemap(oldSitemap)
+   *
+   * @param {sitemapUrlSet} oldSitemap - A sitemapUrlSet object for the old sitemap
+   * @returns {sitemapUrlSet} A sitemapUrlSet object with imported timestamps from the old sitemap.
+   */
+  importTimestampsFromOldSitemap(oldSitemap) {
+    for (url in this.urlSet) {
+      if (url in oldSitemap.urlSet) {
+        var oldUrl = oldSitemap.urlSet[url];
+        var newUrl = this.urlSet[url];
+        if (oldUrl['hash'] === newUrl['hash']) {
+          newUrl['lastMod'] = Math.min(newUrl['lastMod'], oldUrl['lastMod']);
+        }
+      }
+    }
+    return this;
+  }
+}
+
+/**
+ * Returns a md5 hash representation of the content.
+ * Will remove script tags and other HTML before generating the hash.
+ *
+ * @param {string} content - HTML content of the page.
+ * @returns {string} md5 hash representation of the content of the page
+ */
+function getHashOfContent(content) {
+  content = content.replace(/<script\b[^<]*(?:(?!<\/script>)<[^<]*)*<\/script>/gi, '');
+  var $ = cheerio.load(content);
+  var bodyText = $('body').text();
+  var cryptoHash = crypto.createHash('md5').update(bodyText).digest('hex');
+  return cryptoHash;
+}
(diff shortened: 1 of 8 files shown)


Ken-Lo:
Hey @english @jackfranklin 
Would be great to get some thoughts or feedback on this! Sorry, it's failing in CircleCI atm (Cannot find module 'cheerio')... not sure why this is or if I'm supposed to add something different to package.json?
Thanks!

jackfranklin:
@Ken-Lo when you `npm install` something, you also need to regenerate the shrinkwrap with `npm shrinkwrap` (`npm install` will use the shrinkwrap by default)

app/static-build/sitemap.js
```diff
@@ -0,0 +1,153 @@
...
+
+  /**
+   * Loads in an existing XML Sitemap and returns a SitemapUrlSet object.
+   * e.g. var currentSitemap = new SitemapUrlSet('http://example.com').fromXml(xmlFileContents)
+   *
+   * @param {string} xml - Sitemap in XML format
+   * @returns {SitemapUrlSet} A SitemapUrlSet object representing the sitemap.
+   */
+  fromXml(xml) {
+    var thisSitemap = this;
```
lawrencejones: All these `var`s could be `let`s

app/static-build/sitemap.js
```diff
@@ -0,0 +1,153 @@
...
+        var hash = $(this).find('content\\:hash').text();
+        var lastMod = Date.parse($(this).find('lastmod').text());
+        thisSitemap.urlSet[pathName] = {
+          hash: hash,
+          lastMod: lastMod,
+        };
+      }
+    });
+    return thisSitemap;
+  }
```
jackfranklin: throughout this file I'd swap any `var`s for `const` (if you're not going to mutate or reassign), and `let` if you are going to  mutate

app/static-build/sitemap.js
```diff
@@ -0,0 +1,153 @@
...
+   * e.g. var currentSitemap = new SitemapUrlSet('http://example.com').fromXml(xmlFileContents)
+   *
+   * @param {string} xml - Sitemap in XML format
+   * @returns {SitemapUrlSet} A SitemapUrlSet object representing the sitemap.
+   */
+  fromXml(xml) {
+    var thisSitemap = this;
+    var $ = cheerio.load(xml);
+    var sitemapUrls = $('url');
+    sitemapUrls.each(function() {
```
lawrencejones: `function() {...}` -> `() => {...}`

app/static-build/sitemap.js
```diff
@@ -0,0 +1,153 @@
...
+      if (location) {
+        var pathName = url.parse(location).pathname;
+        var hash = $(this).find('content\\:hash').text();
+        var lastMod = Date.parse($(this).find('lastmod').text());
+        thisSitemap.urlSet[pathName] = {
+          hash: hash,
+          lastMod: lastMod,
+        };
+      }
+    });
```
jackfranklin: think this would be clearer if you did `const currentUrl = $(this)`

jackfranklin: if you do

```js
sitemapUrls.each(function() {
  ...
}.bind(this)
```

That ensures the function is called with the scope that it's defined in, and hence you can ditch the `thisSitemap = this`.

script/dump-html
```diff
@@ -11,13 +15,37 @@ if (!serverUrl || !outDir) {
   process.exit(1);
 }
 
+// Download the old sitemap
+// TODO: Download this from live (https://gocardless.com/sitemap.xml)
+// TODO: Initalise SitemapUrlSet with domain name from environmental variable
+https.get('https://dl.dropboxusercontent.com/s/otbggbop9dohag0/sitemap.xml', function(res) {
```
jackfranklin: let's update this before merging!

script/dump-html
```diff
@@ -11,13 +15,37 @@ if (!serverUrl || !outDir) {
...
+  if (res.statusCode == 200) {
+    res.on('data', function(oldSitemapXml) {
+      const oldSitemap = new SitemapUrlSet('https://gocardless.com').fromXml(oldSitemapXml);
+    });
+  }
+}).on('error', function(e) {
+  console.error("Wasn't able to download the old sitemap.");
```
jackfranklin: you can use arrow functions throughout this

also worth in the error call back including the error message:

```js
.on('error', (e) => {
  console.error(`Wasn't able to download old sitemap: ${e.message}`)
});
```

script/dump-html
```diff
@@ -11,13 +15,37 @@ if (!serverUrl || !outDir) {
...
+})
+
+// Is there a better way than using a global variable?
+global.newSitemap = new SitemapUrlSet('https://gocardless.com');
```
jackfranklin: why the need to do this over just:

```
let newSitemap = ...
```

?

jackfranklin:
@Ken-Lo this looks really good! Got some stylistic ES6 comments, let me know if you want to chat through any. Will try to find some time to pull this down locally too.

Ken-Lo:
@jackfranklin Thanks! When I run `npm shrinkwrap`, it changes all of the 'from' values in npm-shrinkwrap.json e.g.
```
      "from": "aws-sdk@2.1.29",
```
becomes:
```
      "from": "https://registry.npmjs.org/aws-sdk/-/aws-sdk-2.1.29.tgz",
```
Is this a problem?

Will go through and fix the other comments as well - thanks!

script/dump-html
```diff
@@ -11,13 +15,37 @@ if (!serverUrl || !outDir) {
   process.exit(1);
 }
 
+// Download the old sitemap
+// TODO: Download this from live (https://gocardless.com/sitemap.xml)
+// TODO: Initalise SitemapUrlSet with domain name from environmental variable
+https.get('https://dl.dropboxusercontent.com/s/otbggbop9dohag0/sitemap.xml', function(res) {
+  if (res.statusCode == 200) {
```
english: Its good to default to using the triple equals (`===`) http://stackoverflow.com/questions/523643/difference-between-and-in-javascript

script/dump-html
english: [`superagent`](https://www.npmjs.com/package/superagent) has a much nicer api than `https`, we use it in https://github.com/gocardless/splash-pages/blob/output-xml-sitemap-v2/app/static-build/builder.js too.

script/dump-html
```diff
@@ -11,13 +15,37 @@ if (!serverUrl || !outDir) {
...
+    res.on('data', (oldSitemapXml) => {
```
english: Another reason to use `superagent` is because the `res.on('data')` event only gives you 'chunks', not the full response. You may get the whole file in one chunk, but I'm not sure what the rules are on when you do and when you don't.

script/dump-html
```diff
@@ -11,13 +15,37 @@ if (!serverUrl || !outDir) {
...
+  console.error(`Wasn't able to download the old sitemap: ${e.message}`);
+});
+
+// Is there a better way than using a global variable?
+global.newSitemap = new SitemapUrlSet('https://gocardless.com');
+
 const processPage = function(pagePath) {
-  return fetchPage(serverUrl, pagePath)
-    .then(writePage.bind(null, outDir))
+  // TODO: Return a promise for both of these functions
```
english: I'd write this as:

```js
  return fetchPage(serverUrl, pagePath)
    .then(function([url, contents], reject) {
      global.newSitemap.addUrl(url, contents);
      return writePage(outDir, [url, contents]);
     })
    .catch(console.error.bind(console));
};
```

Should be a bit easier to read.

script/dump-html
```diff
@@ -11,13 +15,37 @@ if (!serverUrl || !outDir) {
...
+    res.on('data', (oldSitemapXml) => {
+      const oldSitemap = new SitemapUrlSet('https://gocardless.com').fromXml(oldSitemapXml);
+    });
+  }
+}).on('error', (e) => {
+  console.error(`Wasn't able to download the old sitemap: ${e.message}`);
+});
+
+// Is there a better way than using a global variable?
+global.newSitemap = new SitemapUrlSet('https://gocardless.com');
```
english: I don't think you need to set this on `global`? Could you just do `const newSitemap = new SitemapUrlSet('https://gocardless.com');` and refer to `newSitemap` further down in the code?

english: We could maybe take a `sitemapUrl` argument to the script? That way we can run it in staging, or anywhere else and check it works.

app/static-build/sitemap.js
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Returns a md5 hash representation of the content.
+ * Will remove script tags and other HTML before generating the hash.
+ *
+ * @param {string} content - HTML content of the page.
+ * @returns {string} md5 hash representation of the content of the page
+ */
+function getHashOfContent(content) {
+  content = content.replace(/<script\b[^<]*(?:(?!<\/script>)<[^<]*)*<\/script>/gi, '');
```
english: Are you able to use `cheerio` to remove `script`s?

english: Also, do we need to remove them? Or will `$('body').text()` include script content?

jackfranklin: You could `cheerio.load(thing).$('script').remove()` I think

script/dump-html
jackfranklin: :+1: for consistency

app/static-build/sitemap.js
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  /**
+   * Extracts a URL from the sitemap.
+   * e.g. const aboutPage = sitemap.getUrl('http://www.example.com/about/')
+   *
+   * @param {string} pageUrl - URL of the page to add to the sitemap
+   * @return {{Object|boolean}} Returns URL info if exists in the sitemap; otherwise returns false
+   */
+  getUrl(pageUrl) {
```
english: does this function get used anywhere?

english: Ignore that, I'm guessing you're about to use it!

english:
Looking really good so far!

Feel free to come and grab one of us if you need a hand, or if you're ready to get it finished off.

app/static-build/sitemap.js
Ken-Lo: Unfortunately, `$('body').text()` contains the content of the `<script>` tag. Not sure if this a bug or intended behaviour :( Will try Jack's thing to remove it :+1:

Ken-Lo:
Hey guys,
Thanks for the review and for all the comments on the PR!
@english I've finished this off following our discussion earlier in the afternoon. Have also tested `npm run build-staging` and `npm run build-production` locally and verified it works. Anything else or does this look OK to merge?:)
Thanks!

script/dump-html
```diff
@@ -1,23 +1,41 @@
...
 import availableLocales from '../config/available-locales';
 import { getAllPaths } from '../app/router/route-helpers';
 
-const [bin, script, serverUrl, outDir] = process.argv;
+const [bin, script, serverUrl, outDir, target] = process.argv;
+
+const sitemapDomain = target;
+const oldSitemapUrl = `${target}/sitemap.xml`;
 
 if (!serverUrl || !outDir) {
```
english: I missed this yesterday, but we should add `target` here too, so:

```js
if (!serverUrl || !outDir || !target) {
  console.error(`Usage: ${bin} ${script} host directory target`);
  process.exit(1);
}
```

app/static-build/sitemap.js
```diff
@@ -0,0 +1,180 @@
...
+}
+
+/**
+ * Creates a new SitemapUrlSet object.
+ * e.g. let sitemap = new SitemapUrlSet('http://example.com')
+ *
+ * @class
+ * @param {string} domain - Domain to which the sitemap refers e.g. https://example.com
+ */
+export class SitemapUrlSet {
```
english: I would just call this class `Sitemap`. I don't this `UrlSet` adds anything.

script/dump-html
lawrencejones: I would usually use something like commander.js for this. I know it's a really simple use case but you then get nice command line help for free, and it prevents this from ever becoming a script that takes ten positional arguments that no one can remember.

app/static-build/sitemap.js
```diff
@@ -0,0 +1,180 @@
...
+
+  /**
+   * Imports the timestamps from an old sitemap when the content is unchanged.
+   * e.g. newSitemap.importTimestampsFromOldSitemap(oldSitemap)
+   *
+   * @param {SitemapUrlSet} oldSitemap - A SitemapUrlSet object for the old sitemap
+   * @returns {SitemapUrlSet} A SitemapUrlSet object with imported timestamps from the old sitemap.
+   */
+  importTimestampsFromOldSitemap(oldSitemap) {
+    for (url in this.urlSet) {
```
english: @jackfranklin tell me if I'm wrong, but I believe we tend to avoid `for in..` in our JS.

What about something like:

```js
  importTimestampsFromOldSitemap(oldSitemap) {
    Object.keys(this.urlSet)
      .filter((url) => oldSitemap.urlSet[url])
      .filter((url) => oldSitemap.urlSet[url].hash === this.urlSet[url])
      .forEach((url) => {
        const oldUrl = oldSitemap.urlSet[url];
        const newUrl = this.urlSet[url];

        newUrl.lastMod = Math.min(newUrl.lastMod, oldUrl.lastMod);
      });

    return this;
  }
```

This splits up each operation into its own thing, hopefully making it more readable.

english:
A couple final comments, but I'm really happy with it! Merge when you're ready :)

app/static-build/sitemap.js
jackfranklin: :+1:

jackfranklin:
@Ken-Lo as Jamie says, top stuff! After last comments, go for it!

script/dump-html
Ken-Lo: @lawrencejones Thanks! commander.js looks quite handy - do you think it's still useful in this case as all the arguments need to be provided? I haven't changed this in the PR here but could be something to consider later on :)

Ken-Lo:
Thanks guys - very much appreciate all of the help, guidance & comments! Have tidied it up following the final comments above and now going to merge into dev.
Thanks again!

english:
:dancers:

PR #256, opened 2015-07-28
Adding stories to France

adeclosset:
Work in progress : just added the structure. Now need to add the content, which will need vetting from French Talents & 1001 Menus.

Not in scope: Translating to English.

FYI - @Ken-Lo @octaveauger


adeclosset:
@Ken-Lo @octaveauger - Not sure why this keeps failing on the server, since it's not failing locally. However, the footer is duplicated for some reason.
screen shot 2015-07-28 at 19 49 22

app/router/routes.js
```diff
@@ -655,9 +657,12 @@ export const config = Immutable.fromJS([
     },
   ],
   [Stories, { name: 'stories' }, {
-      'en-GB': {
+      en: {
```
Ken-Lo: Best to change this back to 'en-GB' (otherwise you get it appearing on en-IE and en-EU... this is why the test is currently failing)

app/router/routes.js
```diff
@@ -708,4 +713,16 @@ export const config = Immutable.fromJS([
       },
     },
   ],
+  [Stories1001Menus, { name: 'stories_1001_menus', category: 'stories' }, {
+      fr: {
+        path: '/stories/1001-menus',
```
Ken-Lo: I assume this is changing to /references/1001-menus?

app/pages/stories/stories.js
```diff
@@ -1,146 +1,18 @@
...
-                  "We moved to GoCardless as collections can be managed by one person instead of a whole department.
-                  It's great to work with another financial technology company who are committed to helping businesses."
-                </p>
-              </div>
-            </Tabs>
-          </div>
-        </div>
-        <hr className='u-margin-An' />
-        <StartTakingPaymentsCTA />
+      <Page>
```
Ken-Lo: The reason why you're getting two footers is because the `<Page>` component from `<StoriesEn>` and `<StoriesFr>` is nested within the `<Page>` component here. I would move the `isInverted={true}` attribute to here and then skip it out in the other two files :)

Ken-Lo:
Thanks @adeclosset... Nice one! Comments above on the routings, error messages & repeated footers.

I was chatting with @octaveauger earlier this afternoon. I was saying there's a lot of content we write in French first before translating it in to English. Following review from marketing/etc, a significant number of changes is often made in the English version. There is then a tedious process of comparing the diff and translating all the changes back into French.

With regards to these 2 customer stories, do you think we'll want to have them on the English site as well? Not sure how you guys feel about this but it might be more efficient to first write them in English before later translating into French. Saying this, if we're not putting the content on the English splash pages, then it probably isn't necessary to go through this step!

octaveauger:
@adeclosset just need the pic and we're good to go

octaveauger:
@adeclosset I'm merging now

PR #261, opened 2015-08-03
Use <select defaultValue> on lead capture form

Ken-Lo:
The build script currently gives the following warning:
```
Warning: Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>.
Warning: Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>.
```

This is because [in React we should use](https://facebook.github.io/react/docs/forms.html) `<select defaultValue>` instead of `<option selected>`.


Ken-Lo:
Hey @timrogers 
This concerns your "number of payments" drop down on the Contact Sales form. Any chance you'd be able to review?:)
Thanks!

PR #265, opened 2015-08-04
Add Zuora section to the partners page.

swillis:
Zuora have been dragging their feet in getting us the quote, but we want to show them off as a partner. Therefore we have added them _sans-quote_ for the time being in the hope that it will be with us soon.

screen shot 2015-08-04 at 17 08 50


swillis:
@jamesshedden can I get a :+1: ? :smile:

jamesshedden:
@swillis Top work matey :+1:

PR #266, opened 2015-08-17
adding global expansion lead role

moyni:
@jesssummerfield @gabrielamatias all good?


moyni:
@jesssummerfield yes! I'll fix that.

PR #267, opened 2015-08-17
removing customer support france job ad

moyni:
@adeclosset @gabrielamatias French customer support job ad removed.


moyni:
@adeclosset @octaveauger - are you both ok with this ad going down? If so, please check & merge!

octaveauger:
@moyni we are no longer removing this ad so let's not merge

PR #269, opened 2015-08-19
fixing xero description

georg911:
xero doesn’t offer recurring payments


georg911:
Can @jacobpgn please review?

PR #272, opened 2015-08-20
Added Zuora to French partner page

octaveauger:
@adeclosset fyi

### DIFF
--- a/app/messages/fr.js
+++ b/app/messages/fr.js
@@ -78,7 +78,7 @@ export default {
     information2: '',
   },
   features: {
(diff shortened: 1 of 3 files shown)


octaveauger:
@harrison I'm not sure why the test failed - "lets you navigate to a page and then swap languages - fail". There is no indication of which page was tested (I think maybe it's /pro but it works fine in localhost). Can you take a look?

harrison:
@octaveauger looking at the failing e2e. Changed two french titles which seemed wrong.. was `GoCardless: title - GoCardless` so removed `GoCardless: `

harrison:
:+1:

PR #283, opened 2015-09-10
Slight pricing change and typo

adeclosset:
@octaveauger - quick fix here

@Ken-Lo - can push to live please?


Ken-Lo:
@adeclosset Nice one!

For reference, in case you or Octave want to push directly to live... after merging to dev, check CircleCI for deployment errors and also check to make sure it all looks good on staging. You can then deploy directly with the following commands:

```
git checkout dev
git pull

git checkout master
git pull

git merge dev --no-ff --no-edit
git push

git checkout dev
```

PR #305, opened 2015-10-13
Agg `lang` param onto signup page for non-english leads

petehamilton:
Means that we can internationalise the signup page in dashboard for people we know are coming from translated splash pages.

Can be merged in now, but won't take effect until https://github.com/gocardless/enterprise-dashboard/pull/1343 has gone in.


petehamilton:
@greysteil can you review?

greysteil:
:+1:

PR #307, opened 2015-10-15
Remove nonsense meta descriptions & extra-long title on About page

Ken-Lo:
* The title as defined in the message file should be no more than 42 characters long (this is so the HTML `<title>` is [less than 55 characters](https://moz.com/learn/seo/title-tag)).
* A [meta description](https://moz.com/learn/seo/meta-description) should either be well-defined (descriptive & around 155 characters) or left blank.

### DIFF
--- a/app/messages/fr.js
+++ b/app/messages/fr.js
@@ -187,7 +187,7 @@ export default {
     link: 'fr home',
   },
   about: {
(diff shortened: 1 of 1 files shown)


Ken-Lo:
@octaveauger FYI

PR #310, opened 2015-10-16
In the country selector, adjusting position of flag labels and their corresponding svg's.

swillis:
The country labels are now vertically aligned correctly, both in the popover and in the `selected country` component.

screen shot 2015-10-16 at 14 05 11

### DIFF
--- a/app/css/components/checkmark-icon.scss
+++ b/app/css/components/checkmark-icon.scss
@@ -1,4 +1,4 @@
 .checkmark-icon {
   position: relative;
-  top: 3px;
+  top: 2px;
 }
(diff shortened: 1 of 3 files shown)


clippy:
**Changing existing CSS?**

Make sure you know that your changes to existing CSS aren't having any adverse or unexpected consequences across the app you're working with - this should involve tracking down *any* instance of the class, tag or otherwise that you're changing and checking it out locally to make sure everything looks :ok_hand:.

The way things appear affects how GoCardless is perceived - think more "Tech? Internet? We got this!" and less "Your money is safe with us, probably".

swillis:
@jamesshedden can I get a :+1: if you are happy?
:goat:

jamesshedden:
@swillis Hero :+1:

PR #318, opened 2015-10-21
Remove eslint comment from job page

maxmurdoch:
An eslint enable comment got into the splash pages because it was inside the return statement of a call to `render`.

@waltfy cool?
props to @mcollier000 for pointing this out


waltfy:
@maxmurdoch :rocket:

PR #328, opened 2015-11-02
Add Swedish splash pages

duncanbarrigan:
Adding English-language splash pages for Sweden. Starting point was the Irish site, with modifications as required. Tutorials are live with the required changes for sign-up, but I have removed the FAQs for now as they are not all correct for Sweden.

@Ken-Lo could you review please and add a second button at the top of the home page to link to the contact sales form?

Also is there a way of adding a new paragraph to the top of the contact sales form? At the moment it is too crowded. There's no html in the messages files though so I couldn't see how to change this. If it's really painful perhaps we add another paragraph below the hero box?

image


Ken-Lo:
Hey @duncanbarrigan - sorry, I accidentally renamed the branch to Add-sweden-page (case-sensitive) so have had to set up a new pull request here: https://github.com/gocardless/splash-pages/pull/331

PR #340, opened 2015-11-19
Update to data engineer job spec

nhockham:
Following convo with Norberto, we've tweaked the job ad to include requirements for in-depth knowledge of system internals. Also, tried to make it appeal to more senior engineers.


nhockham:
@jesssummerfield pls have a quick look?

@nlopes FYI

jesssummerfield:
@nhockham looks good!

PR #348, opened 2015-12-01
Correcting error in support ad

jesssummerfield:
@harrison this ok? I only changed one letter...

fyi @yoshiotakeuchi


harrison:
:+1:

PR #355, opened 2015-12-17
Get rid of weird Swedish contact sales messaging 🇸🇪

timrogers:
This looked really strange, Autogiro is no longer free, and I'm not convinced the contact sales page is the right place to communicate that we're in beta in Sweden, especially with open signups!

### DIFF
--- a/app/messages/en-SE.js
+++ b/app/messages/en-SE.js
@@ -11,7 +11,6 @@ export default {
     has_percentage_pricing: true,
   },
   prospect_form: {
-    intro_message: 'We are currently beta testing our Autogiro service. Participants can collect Autogiro payments for free during the trial.',
     name_placeholder: 'William Johansson',
     email_placeholder: 'example@companyname.se',
     phone_placeholder: '0844 680 379',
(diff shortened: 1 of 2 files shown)


timrogers:
@harrison A quick one for you.

harrison:
Cool, @duncanbarrigan make sense?

timrogers:
Had a verbal :+1: from @duncanbarrigan

duncanbarrigan:
Makes sense to remove the mention of beta from contact sales and put it somewhere else now that we are opening signups. It should still be visible somewhere to new people though!

The free payments is still an offer through channels like start-up acelerators but don't want it on the website any more - we have some big customers for post-beta in the pipeline and I don't want them to see it!

harrison:
@duncanbarrigan makes sense! You looking at that? Splash pages still not working for you? Can sort this out today.

nagirrabs:
I'm off for the afternoon actually and have a Fintech event tomorrow morning!

Where is it easy to add the beta trial mention from a technical perspective? I'd be happy to add to e.g. the sign-up flow somewhere. Can take a look tonight or tomorrow.

I think my splash pages work fine, but I still have problems with node and running the guides. Would be great to fix that tomorrow afternoon before everyone is off for Xmas if that's possible

nagirrabs:
It seems I managed to comment from my personal account there haha - oops

Issue #351, opened 2015-12-07
Define constants for number of customers, number of countries, etc.


Ken-Lo:
Throughout the splash pages, there are several numbers we refer to a lot e.g. # of customers, # of countries supported, etc. Rather than having to go through the every individual page, it would be good to define some global constants (this would save us having to define it in en.js, fr.js, de.js, es.js and in every single language)/

I'm thinking it would probably be easiest if we define a new messages file for non-localised values. This would sit one-level above en.js, fr.js, etc. Something like:

* global.js (number of customers e.g. 13,000)
  * en.js
    * en-GB.js
  * fr.js (can overwrite based on locale e.g. 13.000)
    * fr-FR.js

Ken-Lo:
@harrison Any thoughts on the above or on the best way to do this?
/cc @octaveauger

harrison:
ah yeah, maybe add `shared.js` to `app/messages` and add it to the merge here https://github.com/gocardless/splash-pages/blob/dev/config/messages.js#L33

Ken-Lo:
See https://github.com/gocardless/splash-pages/pull/357

PR #357, opened 2015-12-17
Add a shared messages file & use consistent # of merchants across the website

Ken-Lo:
This PR:
* Adds a new message file that is shared across all locales (see https://github.com/gocardless/splash-pages/issues/351)
* Defines a `number_of_merchants` message so we're always showing the same number across the whole website (and we don't need to change it in a million places)

### DIFF
--- a/app/messages/de-DE.js
+++ b/app/messages/de-DE.js
@@ -1,5 +1,3 @@
-
-
 export default {
   country: 'Deutschland',
   country_properties: {
--- a/app/messages/de.js
+++ b/app/messages/de.js
@@ -1,5 +1,3 @@
-
-
 export default {
   header: {
     our_products: 'Produkte',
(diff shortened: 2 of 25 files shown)


Ken-Lo:
@harrison Would be great to get your thoughts!
A little annoying having to also define `number_of_merchants` in `fr.js` and `es.js` (this is due to the thousands separator being different in these places). Would be cool to add sprintf-like formatting to `<Message>` but this would probably be over-engineered as we're only using it quite rarely.
Thanks!

harrison:
This feels like a nice improvement! Would be cool to format the number but let's add it when we have more numbers to format. 

Looks like there are some lint errors, :+1: after fixing

Ken-Lo:
Cheers @harrison! Fixed the lint errors and merging!

PR #365, opened 2016-01-06
Update content marketing ad

jesssummerfield:
@petehamilton is this ok?

### DIFF
--- a/app/messages/en.js
+++ b/app/messages/en.js
@@ -508,8 +508,8 @@ export default {
     title: 'Business Development: Germany',
     description: '',
   },
-  jobs_content_community_manager: {
-    title: 'Content & Community Manager',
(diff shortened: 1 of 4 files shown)


Ken-Lo:
@jesssummerfield Could we kindly add a redirect to config/redirects.json? This will redirect any one will goes to the old URL over to the new one :-)
Thanks!

Ken-Lo:
Thanks @jesssummerfield! Small change: in redirects.json, the first value on each line requires /index.html at the end (otherwise, the redirect doesn't work and the old page will remain online!)

Pete - are you able to make this change and check over the PR tomorrow? If not, happy to take a look when I get back from holiday!

Thanks both!

Ken-Lo:
:+1: Merging this in!

PR #369, opened 2016-01-14
Update wording on intl pages

Ken-Lo:


Ken-Lo:
As discussed with @Jennabrown

PR #411, opened 2016-02-10
Sweden edits

duncanbarrigan:
@Ken-Lo : This is a set of edits to the Swedish splash pages. It's mostly minor fixes but I needed to do a review before we send them for translation.

I had some difficulties getting the javascript to load on the FAQ pages. I tried adding the import... line but I think this is why it's failing the CircleCI. Could you take a look please?

If it's hard to fix perhaps it's worth isolating the problem and merging everything that does work ok for now.

### DIFF
--- a/app/messages/en-SE.js
+++ b/app/messages/en-SE.js
@@ -16,16 +16,16 @@ export default {
   },
   hero: {
     header: 'Accept recurring payments across Europe',
-    desc: 'GoCardless allows you to collect Direct Debit (Autogiro) payments from Sweden, the Eurozone and the UK in a single integration.',
+    desc: 'GoCardless allows you to collect Autogiro (Direct Debit) payments from Sweden, the Eurozone and the UK in a single integration.',
   },
   pricing: {
     cost_cap: '20 kr',
     description: 'Collect Direct Debit payments online with fees of just 1%, capped at 20 kr. Scale pricing is available for larger organisations.',
-    pro_cost_per_transaction: '0,50 - 6 kr',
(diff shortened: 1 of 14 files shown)


Ken-Lo:
Hey Duncan - great stuff! 
Just copying in @anothertompetty as I think we're supposed to have a content freeze at the moment on the Features & Pro page. @anothertompetty: OK to update these pages for Sweden or need to branch off `pro-splash-page-updates` instead?
Cheers

anothertompetty:
hey chaps,

yeah @Ken-Lo a good solution would be to either pause this until after we launch the new splash pages, or branch off of `pro-splash-page-updates` and make the changes there. I can see some instances where this will cause conflicts (`features.en.js, pro.en.js`), but the FAQ stuff should be fairly safe.

If it's urgent and we can't hold off, might be best to commit only the changes to the FAQ pages

duncanbarrigan:
Hey guys,

The main reason for this is that we're getting it translated (starting next week probably). I suppose I'll send them the files anyway so it doesn't necessarily have to be live.

The one exception is the messages file. I'm lowering the top Autogiro pricing to 5kr not 6kr and would like that to go live. I also changed the '>500 payments' because otherwise the Pricing page looks horrible with the boxes the wrong size.

The FAQ pages are actually the least urgent, although I would highly recommend we change the customer FAQ for all of the English pages to not say they can log in to their account(!).

samjewell:
Hell yeah - I've wanted these for ages...!

harrison:
@duncanbarrigan should we close this?

duncanbarrigan:
Yeah let's just close it

PR #423, opened 2016-02-22
Only show error message if there’s an error

maxmurdoch:
Steve noticed an empty error message box floating in the Aus landing page. This div will have been on the page for a while but has only displayed since I changed its styling.


jamesshedden:
@maxmurdoch Looks good! :+1: Is it possible this has cropped up anywhere else?

maxmurdoch:
The prospect form actually handles two forms: a prospect form and a holding form. The behaviour is mostly the same: a user fills in a form, sends it to the inbound team, then we display a success or error message). I recently changed the notification messages to be more clear and marry better with our current styling. This means that div above, which will have shown for ages but been kind of invisible, are now more obvious.

PR #453, opened 2016-03-10
Add link to /partners to the More dropdown.

swillis:
The previous splash page update removed /partners from the Nav entirely, this brings it back.

screen shot 2016-03-10 at 13 45 29


swillis:
@jamesshedden quick review if you could squire :smile:

PR #463, opened 2016-03-17
Update the link to the press pack. [Do Not Merge!]

swillis:
Link to the press-pack on `/about` was broken. I've replaced the link, moved the press-pack folder and shared it with the relevant people responsible for updating it.


swillis:
@jamesshedden quick :+1: when you get a mo?

jamesshedden:
@swillis Nice :+1:

We should almost definitely store the press pack URL somewhere so that we can reuse it using our components & messages, rather than manually duplicating it - would make this PR extra sweet if you fancied taking care of that now...

PR #465, opened 2016-03-19
Remove CS ad and updates content manager

gabrielamatias:
- Remove Customer support ad
- updates content manager intro
- changes "The ideal candidate" to "What we're looking for" in ads for consistency

@jacobpgn , @jesssummerfield : please review.


jacobpgn:
There are a few different things happening here - it's best to keep them self-contained so this could be split out into separate pull requests, one for each distinct change :sparkles: 

That said, :+1:

PR #471, opened 2016-03-24
Use messages instead of hardcoding prices

Ken-Lo:


Ken-Lo:
Fyi @duncanbarrigan - this fixes the hardcoded prices.

PR #483, opened 2016-04-08
removing info on pre-populating payment pages

georg911:
as the new API doesn’t support this yet


georg911:
Removing on all languages, EN, DE and FR
@jacobpgn Can you please review? Thx 🐱

jacobpgn:
:+1: Can't wait to add this back in! cc @greysteil

PR #489, opened 2016-04-14
Update local authorities logos.

swillis:
The logos needed to represent local authorities that use GC. Now they do!

screen shot 2016-04-15 at 12 01 53


swillis:
@jamesshedden quick look when you get a mo?

PR #502, opened 2016-04-21
Add Jadu logo to the partner logos image.

swillis:
Jadu have just signed as partner! So I added their logo to the partners page.

image


swillis:
@jamesshedden — super quick look when you get a sec?

jamesshedden:
@swillis Looks good to moi 🏆

PR #524, opened 2016-05-12
Change background sizing to stop images being squashed

swillis:
The image container is set to 50%, and so changes size depending on the width of the screen.

This lead to slightly squashed images in some cases.

But no longer!

screen shot 2016-05-12 at 10 19 00

cc @jamesshedden


clippy:

- [ ] Chrome
- [ ] IE11
- [ ] Firefox
- [ ] Safari
- [ ] iPad

swillis:
@jamesshedden quick look when you get a mo?

PR #525, opened 2016-05-13
updated number of customers for PicDrop

jutta-frieden:
Updated the number of customers from 12k to 16k to reflect most recent numbers.


jutta-frieden:
screen shot 2016-05-13 at 10 54 52

PR #532, opened 2016-05-25
Remove french support ad

jesssummerfield:
@jacobpgn this ok?


jesssummerfield:
@jacobpgn ignore this - hmac checked!

PR #538, opened 2016-06-02
removed text on Mixpanel from German Privacy Policy

jutta-frieden:


jutta-frieden:
@hmarr Thanks for the hint!

PR #547, opened 2016-06-22
Fix flaky specs in home.e2e

harrison:
The tests are failing because the links are trigger full page reloads because the js hasn't managed to
load at the time when the tests run.

We had some expectations which checked the URL of the new page before it was loaded.

There doesn't seem to be a way to wait for a url change to happen which is sad.

This changes the expecations to wait for some text to appear on the page.


harrison:
@waltfy FYI

PR #548, opened 2016-06-22
Stop installing chrome on circle and update node to 6.1.0

harrison:
Chrome takes 4 mins to install on circle. Looks like the script is old and install stuff like ruby, which is mad. We should be able to rely on circles chrome now that's fairly up to date.

Also update node to 6.1.0 in nvmrc and circle.


clippy:

- [ ] Chrome
- [ ] IE11
- [ ] Firefox
- [ ] Safari
- [ ] iPad

harrison:
@waltfy also :)

PR #564, opened 2016-07-05
Move the education page to the industries section

swillis:
**What? / Why?**
Move the education page into the industries folder, so that it appears in the ‘industries’ dropdown in the header.

image


swillis:
@jamesshedden reviewski? 🐐

jamesshedden:
@swillis Looks good, any idea what the `_dm` or `DM` suffix refers to? cc @Ken-Lo

jamesshedden:
@Ken-Lo Oh it refers to direct mail? Not a big deal right now but wondering if we need this — it's a bit too unclear as a suffix to make any sense (esp to newcomers to the repo with no knowledge of why these pages would be different, esp as we're now including this one in the main nav), but be good to know reasons behind it to see if there's a better solution

jamesshedden:
@swillis Code is good though 👍

Ken-Lo:
@swillis @jamesshedden Cool! Thanks for this guys! I would definitely remove the `DM` suffix now, as this is being turned into a proper splash page (original intention was for this to be a transient page). I think the header image is also a different colour to what we use on the rest of the website (so worth thinking about if it's now part of the on-site journey). We can rename `DM` to something else but would like to avoid open-sourcing the marketing strategy as much as possible.

PR #577, opened 2016-07-25
New French homepage

Ken-Lo:
screen shot 2016-08-04 at 11 50 42


Ken-Lo:
I've just gone through this with @comegocardless so going to merge it in now.

PR #597, opened 2016-08-08
Optimising images

waltfy:
Applying lossless compression to `jpeg`/`png` images.


waltfy:
@jamesshedden review on this? If we find that it looks bad we can revert it. 👍

jamesshedden:
@waltfy Looks good 😍😍😍

PR #599, opened 2016-08-08
Update headline on the GC Pro page

Ken-Lo:
We thought "Totally control Direct Debit' was a bit weird, so changing this to "Direct Debit for the Digital Age"

screen shot 2016-08-08 at 15 32 54


Ken-Lo:
@duncanbarrigan look good?

/cc @ruddydunx

duncanbarrigan:
👍

PR #600, opened 2016-08-08
Change hero image on stories page

Ken-Lo:
Use a more generic hero image on the Stories page rather than Hotpod Yoga (looks weird behind a Smart Pension quote!)

screen shot 2016-08-08 at 16 01 25


clippy:

- [ ] Chrome
- [ ] IE11
- [ ] Firefox
- [ ] Safari
- [ ] iPad

Ken-Lo:
Hey @swillis 
Would you mind reviewing? We found this generic hero image from the old SaaS landing page - let me know if you have any thoughts?

swillis:
@Ken-Lo — seems like a sensible idea to me!

One bit of feedback I missed on the last PR: could we reduce the length of the quote by Smart Pension? Ideally it should be 2 lines, 3 at the most. At 4 lines it becomes harder to read and pushes the grid of story tiles further below 'the fold'.

Something like `"GoCardless ensures our transactions are 100% accurate, easy to setup and manage for our clients."` would work much better 😄

Ken-Lo:
@swillis Fantastic feedback - thanks for that! Will merge this in and prepare another PR to fix up the headline quotes ASAP.

PR #609, opened 2016-08-11
Add J&S splash page

Ken-Lo:
screen shot 2016-08-11 at 16 25 10


sjnorth:
@Ken-Lo i am really confused by this. where is the actual text that i wrote? this seems to have no mention of j&s at all, or am i missing something here?

PR #618, opened 2016-08-24
Add Impact Hub case study.

swillis:
Adding the first (hopefully of many) Spanish case study to the site.

I've set up a redirect for `es-es/stories` as it currently does not exist. We'll add this when we have >1 case studies for Spain.

impact hub

PR #629, opened 2016-08-30
Use IfLinkExists on Stories page

Ken-Lo:
Rather than using `IfLocale`, use the more semantically correct `IfLinkExists` tag on the footer of the Stories page (it doesn't make sense to only show that link for countries where Bacs is the domestic scheme).

Also remove the Spanish 'Stories' route as that page currently has no content (shows blank page when you click it in the main nav).


Ken-Lo:
@swillis Some fixes to https://github.com/gocardless/splash-pages/pull/618 - would you mind taking a look? Thanks!

Ken-Lo:
@swillis Just going to merge this, but let me know if you have any questions about it!

PR #648, opened 2016-09-15
Fix pricing for Sweden

a-xin:
Our pricing is up to 5 kr per transaction, not 6 kr, changing the page accordingly.
screen shot 2016-09-15 at 10 40 21


a-xin:
Paired with Duncan

PR #650, opened 2016-09-15
FR: remove pricing comparison table from features and pro page

Ken-Lo:
Features Page:

screen shot 2016-09-15 at 16 28 33

Pro Page:

screen shot 2016-09-15 at 16 28 45


Ken-Lo:
@comegocardless Would you mind reviewing?

PR #652, opened 2016-09-16
Refactoring structure to re-use same StoryPageNew component across locales

jamesshedden:
Reverting recent `StoryPageNewDe` + `StoryPageNewEn` structure back to a single `StoryPageNew` component that's shared across locales & uses `Message` components to swap out translatable content.


jamesshedden:
cc @jutta-frieden 🎉

PR #653, opened 2016-09-16
Changing capitalisation of Gastrofix image files

jamesshedden:
Naming issue where images are not found on live site due to be capitalised 😢


jamesshedden:
cc @jutta-frieden Think this should fix the issue on the live site!

PR #658, opened 2016-09-19
Remove .ruby-version

petehamilton:
CircleCI thinks this is a ruby project and installs a really old version
of ruby (which takes forever as it isn't preinstalled).

This project doesn't use any ruby so we're removing this to prevent ruby
installs at all.


petehamilton:
@Ken-Lo this should give you a pretty decent speedup. Let's see...

Ken-Lo:
@petehamilton Amazing stuff - thanks Pete!

PR #665, opened 2016-09-20
Update terms and conditions

swillis:
Update Merchant Agreement, Partner Agreement & Restricted activities in the section of the website.

Also updated the way in which we handle ‘old Ts & Cs’ so that we have an index from which you can review the previous Ts & Cs across the different sections of the of our terms of service.

**Merchant agreement:**

merchant clipped

**Previous terms of service:**

image


clippy:

- [ ] Chrome
- [ ] IE11
- [ ] Firefox
- [ ] Safari
- [ ] iPad

swillis:
@waltfy could you please give this a review when you get some time? 🙌

app/pages/legal/partners/legal-partners.js
```diff
@@ -10,224 +10,176 @@ export default class LegalPartners extends React.Component {
...
-          You are responsible for capturing and securing any Personal Data you need prior to
-          submitting it to us for processing. GoCardless will use the Personal Data for processing
-          the payment transactions requested by your customers. For the purposes of the Contract,
-          'Personal Data' shall have the meaning given to it by the Data Protection Act 1998.
+          You are responsible for capturing and securing any Personal Data you need prior to submitting it to us for processing. GoCardless will use the Personal Data for processing the payment transactions requested by your customers. For the purposes of the Contract, 'Personal Data' shall have the meaning given to it by the Data Protection Act 1998.
         </p>
 
         <h3 className='u-text-heading-light u-color-dark-gray u-margin-Vm u-text-s'>
-          General
+          10. General
```
waltfy: This could be a list? Thinking about consistency here.

app/router/routes.js
```diff
@@ -591,48 +595,108 @@ export const config = Immutable.fromJS([
       },
     },
   ],
-  [LegalOldRestrictionsIndex, { name: 'legal_old_restrictions' }, {
+  [LegalOldTermsIndex, { name: 'legal_old_terms' }, {
```
waltfy: I'd make sure that these are all right under all languages etc.

swillis: So I've followed the convention used for other pages that are only in English — which is to account for all of the `En-##` locales.

Does that sound about right?

image

app/pages/legal/partners/legal-partners.js
swillis: Updated!

app/router/routes.js
waltfy: @swillis 👍 for the picture of code — historic moment

#samspearls

PR #677, opened 2016-09-28
[DNM] Test modelines to improve github highlighting

petehamilton:
See if we can trick github, as per https://github.com/github/linguist#using-emacs-or-vim-modelines

Seems to work...

@harrison kinda gross - but what do you think re: adding modelines to our JS files so github shows diffs nicely (rather than the crazy red we currently get)?

Alternative would be to rename all `.js` -> `.jsx` which I'd kinda prefer... You?


harrison:
@petehamilton yeah would be nice to rename all files to jsx, not sure how painful this would be but feels more robust. I would definitely forget to add that comment to files if adding new ones.

jamesshedden:
@petehamilton @harrison 

Cool idea Pete, nice one for suggesting a way to improve this 🙏 

Feels like it makes sense to close this for now as we've added renaming `js` -> `jsx` as things to look at as part of giving the repo some overdue maintenance (which we're now doing & continuing in Jan) — shout if any problems or if isn't right to close this 👍

PR #679, opened 2016-09-29
fix german typos

georg911:
Fixing "autorisiert" as seen yesterday.

@jutta-frieden can you review please?


georg911:
@jutta-frieden updated! another looksie?

jutta-frieden:
Looking great, go for it! @georg911

PR #691, opened 2016-10-14
Replace swedish number with UK number on Swedish splash pages

duncanbarrigan:
Removing the public Swedish number from our website. I don't have splash-pages running locally so this is a fairly blind PR - can someone check it works please? @a-xin / @amfarrell / @hmac ?


a-xin:
@duncanbarrigan Seems to work, `/en-se/legal` has the UK phone in the sidebar, in the text as well as in the footer. Not sure if there are any other pages to check, though?

## gocardless.com/en-se/legal/
screen shot 2016-10-14 at 17 00 43

Ken-Lo:
@duncanbarrigan Removing this from `en-SE.js` just means we pull in the default UK phone number with IDD (+44 20 7183 8674) from `en.js` - assume that's what you were looking for?

Ken-Lo:
:+1: Will prepare another PR later today to do the same on the guides

PR #693, opened 2016-10-17
Always use ?s= parameter when linking to contact_sales route

Ken-Lo:
The sales team would like better insight on how people get to the Contact Sales page. At the moment, we provide this insight in Salesforce by tracking the Contact Sales URL and `s` parameter.
e.g. `/contact-sales/?s=pricing` or `/contact-sales/?s=pro`.

Unfortunately, we aren't currently using the `s` parameter very consistently. This PR does two things:
- Remove hard-coded query parameters (easy to make mistakes e.g. when copying & pasting into another file)
- Automatically generate the query parameters within the `<Link>` component when `to="contact_sales"` is specified.

### DIFF
--- a/app/components/link/link.js
+++ b/app/components/link/link.js
@@ -1,3 +1,4 @@
+import assign from 'lodash/object/assign';
 import React, { PropTypes, Component } from 'react';
 import Message from '../message/message';
 import Router from 'react-router';
@@ -17,9 +18,19 @@ export default class Link extends Component {
     onClick: PropTypes.func,
   }
 
+  static contextTypes = {
+    routeName: PropTypes.string.isRequired,
+  }
+
   render() {
+    const { routeName } = this.context;
+    const props = assign({}, this.props);
+    if (props.to === 'contact_sales') {
+      props.query = { s: routeName };
+    }
+
     return (
-      <RouterLink {...this.props} activeClassName='is-active'>
+      <RouterLink {...props} activeClassName='is-active'>
         { this.props.pointer &&
           (<Message pointer={this.props.pointer} />) || this.props.children }
       </RouterLink>
--- a/app/pages/partners/partners.de.js
+++ b/app/pages/partners/partners.de.js
@@ -58,8 +58,7 @@ export default class PartnersDe extends React.Component {
                 </a>
               </li>
             </ul>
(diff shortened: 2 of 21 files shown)


Ken-Lo:
Hey @harrison 
Could I please get your thoughts on this?
Thanks! :)

/cc @jmrobins fyi

harrison:
@Ken-Lo hm interesting, not sure about changing Link, ideally we would only change the specific route contact sales instead of changing globally.

Where and how does this param get used? Sent to GTM from params?

Might be a bit tricky with our router config but think we want to do this as a hook when navigating to the contact sales route, when this happens we can look at the previous routes name and add it to the query params.

Ken-Lo:
@harrison Good feedback, thanks! Ideally I would like to scrap the `?s=` parameter as well (it's ugly and causes duplicate URL issues, messy analytics, etc). It's what we currently use however, though we do so inconsistently so the data isn't good. The PR above just ensures we use `?s=` correctly through the website.

To answer your question:

We don't actually need this in GTM or for any digital channels - this is solely for @jmrobins who uses it to track performance in Salesforce (so the contact sales URL contains additional insight e.g. the page they were viewing immediately prior to contacting sales).

Two other options (both requiring more work):
* We could update the `<ProspectForm>` component to send in the referring route name (would need to update backend to pass info through to Salesforce).
* Could use something like `history.replaceState()` to populate this parameter when the contact sales route is viewed? Would still get ugly URLs and there is additional code complexity, so not sure if there's any benefit to this more complicated option?

harrison:
@Ken-Lo hm the stuff we chatted about won't work with the current version of the router, will have a go at updating react and react-router.

Ken-Lo:
Hey @harrison 
Cool - thanks for investigating! Let me know if you have any ideas after updating the router!

harrison:
@Ken-Lo heads up on this, haven't had much time to work on this but the plan is for @jamesshedden and Joe to work on refactoring and updating dependencies when he joins.

We need to give this project some love before we start any big pieces of work :)

@jamesshedden FYI this is about updating react router so we can use some new features in it.

jamesshedden:
@Ken-Lo @harrison Hey, based on chats above, closing this for now as sounds like something we'll want to revisit in Jan once had time to do the necessary maintenance on the repo — shout if any problems :)

PR #696, opened 2016-10-19
Rewriting front end developer job ad

harrison:
Changing from front end designer to front end developer emphasising developer chops more than design skills.


harrison:
@jamesshedden review?

jamesshedden:
@harrison Nice work :+1:

Ken-Lo:
@harrison Nice one - don't forget `redirects.json` as well or the old page also remains online (we currently have hundreds of old job adverts - would be great to delete these at some point)

harrison:
@Ken-Lo nice catch!

PR #703, opened 2016-10-21
changed image files to all lower case

jutta-frieden:
Changed file names to all lower case and removed special German character ü.


jutta-frieden:
@jamesshedden and @a-xin Just fixed the image names (special character probably caused issues with the final image). Want to have a quick look?

PR #706, opened 2016-10-26
Correct links to point to the developer splash page and the API reference as appropriate

timrogers:
This rationalises how we link to developer resources from the splash pages, storing two URLs (`api_reference_link` and `developer_link` in the locale data) and then using them to build links using the `Href` component.

This allows us to, on the French site, correctly point to the French developer splash page at <https://developer.gocardless.com/fr>.

### DIFF
--- a/app/components/footer/footer.js
+++ b/app/components/footer/footer.js
@@ -84,9 +84,9 @@ class Footer extends React.Component {
                 <Link to='faq_pro_dashboard' pointer='faq_pro_dashboard.link_title' className='page-footer__link u-link-invert' />
               </IfLinkExists>
               <li>
-                <a href='https://developer.gocardless.com/' id='track-footer-api-docs' className='page-footer__link u-link-invert'>
+                <Href to='developer_link' id='track-footer-api-docs' className='page-footer__link u-link-invert'>
                   <Message pointer='api_docs.nav_title' />
-                </a>
+                </Href>
               </li>
               <IfLinkExists to='security' tagName='li'>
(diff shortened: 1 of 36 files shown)


timrogers:
@jamesshedden Can you take a look at this? :smile:

app/components/footer/footer.js
```diff
@@ -84,9 +84,9 @@ class Footer extends React.Component {
                 <Link to='faq_pro_dashboard' pointer='faq_pro_dashboard.link_title' className='page-footer__link u-link-invert' />
               </IfLinkExists>
               <li>
-                <a href='https://developer.gocardless.com/' id='track-footer-api-docs' className='page-footer__link u-link-invert'>
+                <Href to='developer_link' id='track-footer-api-docs' className='page-footer__link u-link-invert'>
```
jamesshedden: @timrogers Been a while since I worked in this repo but any reason why these aren't using the `<Link>` component?

jamesshedden: (applies to whole PR where `Href` has been used in place of `Link`)

timrogers: Looked at this myself and the `Link` component is for when you're linking to a route within the site, and `Href` for external links (e.g. in the translations).

Ken-Lo:
Nice one @timrogers!
Might be cleaner to put the default value of `developer_link` in `shared.js` (so this applies to all locales by default), then overwrite it specifically in `fr.js` :-)

PR #724, opened 2016-11-10
Update the options for monthly collections when signing up

timrogers:
We want to better express our ability to handle larger numbers of transactions to show that we cater to larger merchants. This requires changes in the legacy app, the Enterprise Dashboard and the splash pages.

Fixes <https://gocardless.myjetbrains.com/youtrack/issue/UXR-68>.

### DIFF
--- a/app/components/prospect-form/prospect-form.js
+++ b/app/components/prospect-form/prospect-form.js
@@ -173,9 +173,11 @@ export default class ProspectForm extends React.Component {
                       <option value=''>
                         <Message pointer='prospect_form.sales.number_of_payments_placeholder' />
                       </option>
-                      <option value='0-100'>0-100</option>
-                      <option value='100-500'>100-500</option>
-                      <option value='500+'>500+</option>
+                      <option value='0-99'>0-99</option>
+                      <option value='100-999'>100-999</option>
+                      <option value='1000-9999'>1000-9999</option>
+                      <option value='10000-99999'>10000-99999</option>
+                      <option value='100000+'>100000+</option>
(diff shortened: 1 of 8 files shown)


timrogers:
@Ken-Lo Can you take a look at this? :smile:

Ken-Lo:
@timrogers Nice!

Three things:
1) Assume we've already updated the backend code to accept these values?
2) Will need to update the success messages in https://github.com/gocardless/splash-pages/blob/dev/app/messages/en.js#L59 and other locales as well (or we can just refactor to take this out as I think we're always showing the same message anyway).
3) Would be good to double-check if any changes need to made to the guides as well (looking at [this](https://github.com/gocardless/guides/blob/e240e7f7514f662715ce233c0e018208b4b51e04/_assets/js/directives/contact-sales-form-submit-directive.js), I think there shouldn't be?)

timrogers:
1) The backend doesn't need any changes - it doesn't actually validate these values.
2) Looks like we should just refactor those out - I'll look at doing that.
3) Is that even used anywhere? I can't find any references to the directive... (But I also don't know Angular!)

timrogers:
Found it - we do use the form in a few places in the guides, but we don't have the number of payments field :smile:

Ken-Lo:
:+1: Nice one Tim!

RE #1: I was probably thinking of Salesforce. I vaguely remember we had different values in the past on the French website, but they were never passed through or tracked correctly (somewhere down the line, either in GC or SF, the values were lost).

Thanks!

PR #725, opened 2016-11-10
Improve the formatting of monthly collections options

timrogers:
As requested by @nicola-gc.

PR #727, opened 2016-11-14
Move monthly collections options into locale-specific strings

Ken-Lo:
So we can display the # of monthly payments dropdown with the right formatting e.g. `1,000 - 9,999` in UK and `1.000 - 9.999` in EU.


Ken-Lo:
Hey @timrogers 
Just some changes to https://github.com/gocardless/splash-pages/pull/725 to use locale-specific strings. Would you mind reviewing?:)
Thanks!

Ken-Lo:
@timrogers Thanks! I guess we'd ideally like to build something like [Number.toLocaleString()](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) in to the framework - as there are lots of places where we re-define messages with different thousands/decimal separators :)

timrogers:
@Ken-Lo Agreed! We should definitely look at doing that as a refactor.

PR #739, opened 2016-11-24
Redesigning 'Partners' page, moving existing `/partners` to `/partner-with-us`

jamesshedden:
**DNM**

- [x] Align logos
- [x] Add link to new Partners page somewhere in nav
- [ ]  Final naming/copy

---

This PR introduces a new design for our 'Partners' page focused on showcasing the integrations that our merchants can connect with, rather than advertising the benefits of becoming a GoCardless partner — as a result, I've also suggested here moving the existing `/partners` design to a new `/partner-with-us` URL.

image

**The implementation of the new page is based on these principles:**

- The array of partners we offer should be managed via one constants file, via which the content can be more easily amended

- The template must single-handedly reflect the contents of the constants file  — rather than the content being split into multiple templates & routes, which would require maintenance of more than just the constants file when updating partners, categories, etc

**How the content is structured — bit of context to get your head around here so happy to clarify in person**

- There is a list of partners — each partner can belong to one or more categories

- The categories themselves are split into two types. This is just for the purposes of separating 'App' and 'Business' partners in the nav — categories can also have no 'type', if we don't want them to appear in the nav

- Each category can be featured on the home page, showing a selection of partners for that category — this is set on an individual category level with `isFeaturedOnHome`

- That selection of partners is set at an individual partner level — for each partner, we list the categories it should belong to, and whether for each one it should appear as part of the home page selection for that category  — this is set with `isFeaturedOnHome` for each partner's category

- In addition to the home page, each category has its own page, listing all the partners in that category

- Each category page can also have its own featured section, in which case the featured partners in that category are listed first, then the regular partners — otherwise they're all just listed together — this is set with `hasFeaturedSection`

- The partners featured on an individual category page are not necessarily the same as the ones that are included in the previously mentioned selection of featured categories on the home page — so again, this is set on the level of individual partners — we list the categories each partner belongs to, and for each category whether that partner should be featured on the respective individual category page — this is set with `isFeaturedOnCategory` for each partner's category

- Each partner links to another page, which can be an internal route (eg our own page showcasing the partnership) or an external URL (eg the partner's own guide to connecting with GoCardless)


clippy:

- [ ] Chrome
- [ ] IE11
- [ ] Firefox
- [ ] Safari
- [ ] iPad

jamesshedden:
@harrison Hey, got some stuff to attend re: final content but would be awesome to get a pair of eyes on the approach so far 🙏

app/messages/en.js
```diff
@@ -318,9 +318,16 @@ export default {
     explainer: 'Total control over Direct Debit',
   },
   partners: {
+    title: 'Partners',
+    nav_title: 'Partners',
+    link_title: 'Partners',
+    description: 'Partners',
+    explainer: 'Partners',
```
harrison: need this?

app/constants/partners.js
```diff
@@ -0,0 +1,754 @@
...
+      source: '/images/partners/logos/cobot-logo.jpg',
+      width: 120,
+    },
+    link: 'https://www.cobot.me/guides/credit-cards-&-direct-debit',
+  },
+];
+
+export const HOME_NEW_SECTION_ACTIVE = true;
+
+export const sortCategoriesByType = _.map(CATEGORY_TYPES, (type) => {
```
harrison: should this be sorted? or update name?

app/pages/partners/partners.js
```diff
@@ -1,22 +1,275 @@
...
+        <div>
+          <div>
+            <h2 className='u-text-heading u-color-dark-gray
+              u-text-l u-text-light u-inline-block'>
+              { featuredCategory.name !== 'new' ? 'Top ' : null }
+              { featuredCategory.label } partners
+            </h2>
+
+            {
+              featuredCategory.name !== 'new' ?
```
harrison: this new label is a bit weird, why can you not view all the new ones? Also use constant?

app/pages/partners/partners.js
```diff
@@ -1,22 +1,275 @@
...
+              <Link to='partners'
+                className='u-color-primary u-text-upcase u-text-xxs u-link-clean
+                u-text-heading u-text-semi u-margin-Txs u-pull-end'
+                params={{ category: featuredCategory.name }}>
+                View all
+              </Link> : null
+            }
+
```
harrison: 🔪

app/constants/partners.js
```diff
@@ -0,0 +1,754 @@
...
+
+export const presentPartnersFeaturedOnHome = () => {
+  return _.map(getCategoriesFeaturedOnHome(), (category) => {
+    return {
+      name: category.name,
+      label: category.label,
+      partners: getPartnersInHomeFeaturedCategory(category.name),
+    };
+  });
+};
```
harrison: also move this presentation logic to the page? `getCategoriesFeaturedOnHome` is similar to `presentPartnersFeaturedOnHome` and possibly nice to keep all the logic there?

app/pages/partners/partners.js
```diff
@@ -1,22 +1,275 @@
...
+  renderAllPartnersInCategory() {
+    return (
+      <div className="u-margin-Vxl grid">
+        { this.renderPartnerCards(filterPartnersByCategory(this.state.activeCategory.name)) }
+      </div>
+    );
+  }
+
+  getFeaturedPartners() {
+    return this.presentPartners(
```
harrison: not sure you need `presentPartners`, could just inline:

```
_.filter(filterPartnersByCategory(this.state.activeCategory.name), (partner) => {
  return this.isPartnerFeaturedOnCategory(partner);
})
```

app/pages/partners/partners.js
```diff
@@ -1,22 +1,275 @@
...
+                  }
+                </div>
+              </div>
+            </div>
+
+            <hr className='section-divider' />
+
+            <div className='site-container u-text-center u-padding-Vxxl'>
+              <div className='product-grid__container'>
+
```
harrison: 🔪

app/router/routes.js
```diff
@@ -817,6 +819,12 @@ export const config = Immutable.fromJS([
       },
     },
   ],
+  [Partners, { name: 'partners' }, {
+      'en-GB': {
+        path: '/partners/?:category?',
```
harrison: Comment about the question mark weirdness to make it optional?

app/constants/partners.js
```diff
@@ -0,0 +1,754 @@
...
+  },
+  { // 👇 Doesn't have a `type` so doesn't show in sidebar
+    name: CATEGORY_NAMES.NEW,
+    label: 'New & notable',
+    isFeaturedOnHome: true,
+    hasFeaturedSection: false,
+  },
+];
+
+export const PARTNERS = [
```
harrison: not really relevant until we know what we will need but was thinking about localising this, and we'll possibly have to move label to messages. Also imagining we'll have different partners in different countries with different categories so maybe you just have separate ones for different countries, which wouldn't be so bad. Actually maybe rename this to partners.en.js to make it clear for future?

jamesshedden: @harrison Yep makes sense, will rename to `partners.en.js` 👍

app/constants/partners.js
```diff
@@ -0,0 +1,754 @@
...
+    ],
+    logo: {
+      source: '/images/partners/logos/cobot-logo.jpg',
+      width: 120,
+    },
+    link: 'https://www.cobot.me/guides/credit-cards-&-direct-debit',
+  },
+];
+
+export const HOME_NEW_SECTION_ACTIVE = true;
```
jamesshedden: @harrison Good spot!

jamesshedden:
@harrison Thanks for the feedback! 🎱

harrison:
screen shot 2016-11-30 at 18 08 57

@jamesshedden went through some wording with db and trying to keep it in sync with the dashboard. Also fixed an issue where it wouldn't transition when the react app was doing the routing. Let's chat tomorrow morning and figure this out.

jamesshedden:
@harrison Ah ok, nice work! I'm WFH today writing reviews, but Slack me with anything I can help with — otherwise feel free to jump on and ship 🚢

harrison:
screen shot 2016-12-01 at 10 50 57
screen shot 2016-12-01 at 10 50 52
screen shot 2016-12-01 at 10 50 46
screen shot 2016-12-01 at 10 51 02

PR #745, opened 2016-11-28
Update Utilities page to use Bulb Energy quote

Ken-Lo:
Feature Bulb Energy case study on our Utilities landing page.

To do:
- [x] Get longer quote


clippy:

- [ ] Chrome
- [ ] IE11
- [ ] Firefox
- [ ] Safari
- [ ] iPad

PR #755, opened 2016-12-05
Adding routes, messages & rendering for static category pages

jamesshedden:
The site was not previously able to serve static pages for our partner apps categories based on the previous setup — this PR adds routes, messages & some conditional rendering (based on reading URLs rather than React Router params when an individual category route is being rendered) to fix this issue.


jamesshedden:
@harrison Be great to get thoughts on this!

cc @Ken-Lo — this should fix the issue you reported this morning 🌞

Ken-Lo:
@jamesshedden Cool, thanks a lot! Will leave you and @harrison to discuss code but some notes on the messages file:
* I think we can probably take out `nav_title` as that's only used by the main navigation?
* Can we give each of the routes a unique title? e.g. Connect to Accounting Partners, etc.
* If possible, can we also leave the description blank? Happy to optimise these in a later PR, but would be best to leave blank rather than having the same on each route!
Thanks!

app/pages/partners/partners.en.js
```diff
@@ -14,19 +14,34 @@ import {
...
+    super(props, context);
 
     this.state = {
-      activeCategory: getCategory(this.props.params.category),
+      activeCategory: getCategory(this.getCategoryName(this.props.params.category)),
     };
   }
 
+  getCategoryFromUrl() {
+    return this.context.pathname.replace('/partners/', '').replace('/', '');
```
harrison: could regex this, `.replace(/\/|partners/g, '')`

app/messages/en.js
```diff
@@ -322,6 +322,66 @@ export default {
...
+  partners_subscription: {
+    title: 'Connect to our partners',
+    nav_title: 'Connect to our partners',
+    description: 'Use your GoCardless account with a pre-built integration',
+  },
+  partners_utilities: {
+    title: 'Connect to our partners',
+    nav_title: 'Connect to our partners',
+    description: 'Use your GoCardless account with a pre-built integration',
+  },
```
harrison: this is a bit annoying :(

app/pages/partners/partners.en.js
jamesshedden: @harrison Good spot ✅

app/messages/en.js
jamesshedden: @harrison Ah yep, have removed repetition & the `nav_title` property from each one based on @Ken-Lo feedback 👍

jamesshedden:
@Ken-Lo Nice one, thanks for the feedback — taken care of those!

@harrison Thanks for thoughts — yep would be ideal to not have to manually edit routes & messages files as a requirement to edit this section, though not sure what best alternative is 🤔

PR #760, opened 2016-12-07
Fix context pathname

harrison:
Read the next context when props change so it doesn't render the
previous page.


harrison:
slack based 👍 from @jamesshedden

PR #771, opened 2016-12-15
Add missing Irish redirects

Ken-Lo:
Redirects the following two pages from Ireland:
* Old Referral Scheme page
* Old Case Studies page


Ken-Lo:
@sjnorth @jmsgdy This should fix the issue you guys are having.

Fyi @megaslug on redirects as well :)

PR #772, opened 2016-12-16
Add Debitoor landing page

Ken-Lo:
screen shot 2016-12-16 at 17 13 56

screen shot 2016-12-16 at 17 14 13


Ken-Lo:
I've just gone through with with @eobermaier and @carlesmarcos :)

@megaslug Fyi, some new partner pages for Debitoor in Germany and Spain!

PR #777, opened 2016-12-19
[Do Not Merge] Simplified header for case studies page

Ken-Lo:
screen shot 2016-12-19 at 15 08 06

Initial ideas/proof of concept of how this might look.

See https://trello.com/c/Ls768eJZ/96-simplified-header-for-case-studies-page


megaslug:
@Ken-Lo had a look at this with Amir. all looks good.

PR #783, opened 2016-12-22
Fix broken links on SMB funnel

Ken-Lo:
SMB funnel has two broken links:
- `Explore all integrations` should go to `partners` route and not sandbox signup
- `Sign up now` should go the sign up page and not the pricing page!


Ken-Lo:
Fyi @megaslug - I'm going to merge this in as it's a fairly minor update & also fairly critical, but let me know if you have any qs!

PR #787, opened 2016-12-22
Display updated homepage for en-GB, not other en- locales

megaslug:
Display updated homepage to:
- en-GB

Display previous homepage to:
- en-US
- en-SE
- en-NZ
- en-IE
- en-EU
- en-CA
- en-AU


clippy:

- [ ] Chrome
- [ ] IE11
- [ ] Firefox
- [ ] Safari
- [ ] iPad

app/pages/home/home.js
```diff
@@ -20,9 +21,9 @@ export default class Home extends React.Component {
     return (
       <div>
         <IfLocale showHoldingPage={false}>
-          <Translation locales={['fr', 'de', 'es', 'nl']}>
+          <Translation locales={['fr', 'de', 'es', 'nl', 'en-US', 'en-SE', 'en-NZ', 'en-IE', 'en-EU', 'en-CA', 'en-AU']}>
```
Ken-Lo: Can we do:
`<Translation locales={['fr', 'de', 'es', 'nl', 'en']} exclude={['en-GB']}>`

Would rather we blacklist `en-GB` rather than manually whitelist (in case we add or remove locales in future, it gets confusing and we need to update this list in lots of places.

Ken-Lo:
:+1: Looks good! One minor comment about blacklisting/whitelisting then good to go!

PR #791, opened 2016-12-22
Move homepage CTAs to the bottom

Ken-Lo:
After speaking with Nicola and Polly, we've decided to move the CTAs on the homepage back to the bottom.

screen shot 2016-12-22 at 18 09 37


Ken-Lo:
@megaslug Would be great if you could take a look! :-)

megaslug:
@Ken-Lo looks good!

PR #794, opened 2016-12-23
Translate Developers page into French

megaslug:
developers-fr


app/pages/developers/developers.fr.js
```diff
@@ -0,0 +1,432 @@
...
+
+                  <p className='u-color-dark-gray u-margin-Txs'>
+                    Notre équipe de spécialistes API est disponible pour répondre aux questions pendant votre intégration
+                  </p>
+                </div>
+              </div>
+            </div>
+
+            <div className='u-margin-Tm'>
+              <a href='https://developer.gocardless.com/getting-started/api/introduction/'
```
Ken-Lo: https://developer.gocardless.com/fr/commencez-maintenant/api/introduction/

app/pages/developers/developers.fr.js
```diff
@@ -0,0 +1,432 @@
...
+                  </p>
+                </div>
+
+                <div className='u-size-3of4 u-margin-Bl'>
+                  <h3 className='u-text-heading u-color-dark-gray u-text-light u-text-no-smoothing u-text-m'>
+                    Documentation claire <br/>et accessible
+                  </h3>
+
+                  <p className='u-color-dark-gray u-margin-Txs'>
+                    La <a href="http://developer.gocardless.com">documentation</a> complète incluant le guide "pas à pas" .
```
Ken-Lo: Rather than hard-coding, can use <Href to='api_docs.path'>` to automatically pick up localised API docs from message file.

app/pages/developers/developers.en.js
```diff
@@ -0,0 +1,432 @@
...
+import EuropeIcon from '../../icons/svg/globe-europe';
+import FutureProof from '../../icons/svg/FutureProof';
+import PlugIn from '../../icons/svg/PlugIn';
+import ReadPdf from '../../icons/svg/ReadPdf';
+
+export default class DevelopersEn extends React.Component {
+  displayName = 'DevelopersEn'
+
+  render() {
+    const phpCodeBlock = `<code class="code-examples__tabs-code language-php hljs">$subscription = $client-&gt;subscriptions()-&gt;create([
```
Ken-Lo: Can code block variables be placed in `developers.js` as they should be the same in both `<DevelopersEn />` and `<DevelopersFr />`?

Ken-Lo: On second thoughts, if we wanted to translate the comments into French, might be best to keep them separate.

Ken-Lo:
@megaslug Amazing stuff Joe :) Added a few comments/suggestions for the new year!

app/pages/developers/developers.fr.js
megaslug: thanks @Ken-Lo, will get that updated

app/pages/developers/developers.fr.js
megaslug: thanks @Ken-Lo, will get that fixed

app/pages/developers/developers.fr.js
megaslug: This should address the issue: https://github.com/gocardless/splash-pages/pull/795

app/pages/developers/developers.en.js
```diff
@@ -0,0 +1,432 @@
...
+                "The API is great from a technical perspective and using GoCardless means we can avoid the legacy banks payment networks."
+              </p>
+              <p className='u-color-invert u-margin-Bl u-text-s'>Bill Packman, COO</p>
+            </div>
+
+            <h2 className='u-text-heading u-color-dark-gray u-text-light u-margin-Txxl'>
+               Ready to start testing today?
+            </h2>
+
+            <a href='https://manage-sandbox.gocardless.com/signup' className='btn btn--hollow u-margin-Txxl'>
```
Ken-Lo: Hey @megaslug 
Just wondering whether we should maybe move the sandbox signup URL to a pointer and then link to it through `<Href to='' />`? We do actually have a [sandbox signup page here in French](https://manage-sandbox.gocardless.com/signup?lang=fr) so maybe worth linking to that instead as it should boost the conversion :)

app/pages/developers/developers.fr.js
```diff
@@ -0,0 +1,432 @@
...
+          <div className='site-container u-padding-Bxl u-text-center'>
+                <h1 className='u-text-heading u-color-invert u-text-xl u-text-light'>
+                   Construisez mieux, plus intelligemment, <br/>plus rapidement avec notre API
+                </h1>
+
+                <p className='u-text-heading-light u-text-m u-color-invert'>
+                   Focus sur une intégration incroyable du prélèvement automatique, pas sur une intégration de 0.
+                </p>
+
+                <a href='https://manage-sandbox.gocardless.com/signup' className='u-link-color-p u-text-underline'
```
Ken-Lo: Hey @megaslug 
Just wondering whether we should maybe move the sandbox signup URL to a pointer and then link to it through `<Href to='' />`? We do actually have a [sandbox signup page here in French](https://manage-sandbox.gocardless.com/signup?lang=fr) so maybe worth linking to that instead as it should boost the conversion :)

app/pages/developers/developers.fr.js
```diff
@@ -0,0 +1,432 @@
...
+
+        <div className='u-text-center u-padding-Vxl'>
+          <div className='u-padding-Vxl site-container'>
+            <h2 className='u-text-heading u-color-dark-gray u-text-l u-text-light u-margin-Bl'>
+              Rapide à construire avec un ensemble d'outils et des ressources fiables
+            </h2>
+
+            <div className='grid'>
+              <div className='grid__cell u-size-1of2'>
+                <img src='/images/funnels/robust_tools.png' width='454px' />
```
Ken-Lo: Not urgent, but would be great to get a localised screenshot in French if possible, the equivalent FR page [can be found here](https://developer.gocardless.com/fr/commencez-maintenant/api/ajouter-votre-premier-client/).

Also, I thought the [`width` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Specifications) only supports pixels so valid HTML would be `width='454'`?

Thanks!

app/pages/developers/developers.fr.js
megaslug: @Ken-Lo good spot. will update this.

app/pages/developers/developers.fr.js
megaslug: @Ken-Lo would be good to find the design to add this. I'll ask @swillis when he's around.

In terms of width attribute, px is the only supported value. Would be good to add height as well. Although, both of these should sit in CSS, declaring them in HTML reduces content jumps during page load.

PR #801, opened 2016-12-29
Show local phone numbers underneath Contact Sales forms

Ken-Lo:
Show local phone numbers without IDD code (i.e. without the +44 or +33 prefix) underneath Contact Sales forms


Ken-Lo:
@megaslug We'd love to change this on the UK & FR websites as we think it should improve the usability. Would love your thoughts!

megaslug:
@Ken-Lo looks good

PR #802, opened 2016-12-29
Fix /direct-debit/ path proxy

Ken-Lo:
Our current proxies on localhost are a little bit brittle.

**Intended behaviour:**
The `/direct-debit/` folder and `/direct-debit` should be proxied to `guides`, everything else should stay in `splash-pages`.

**Bug:**
With our current proxy settings, `http://localhost:4402/direct-debit-api/` is also proxied to `guides`. This PR resolves the problem by updating the path match patterns.


Ken-Lo:
@harrison @megaslug Would love your thoughts. We could probably get away with only matching `/direct-debit/*` as all of the DD guide content sits in that folder. However, this fails in the edge-case where you visit http://localhost:4402/direct-debit (which works in production when you visit https://gocardless.com/direct-debit).

megaslug:
@Ken-Lo looks good to me, but would be great to hear what @harrison has to say

harrison:
Looks good to me! 👍

PR #807, opened 2016-12-30
Hide en-AU, en-US, pt-BR, en-NZ from footer dropdown

megaslug:


megaslug:
@Ken-Lo alternative to this is using `show_holding_page`, which currently gives you value for the country you are on. To access this property for all available locales I would need to add this perhaps as a separate file or extend render.js with something like:

```
let activeCountries = {};
availableLocales.forEach(function(locale) {
  activeCountries[locale] = localeMessages[locale].country_properties.show_holding_page;
});
```

I thought that since this is a temporary change, having an explicit list is a bit more straightforward.

Let me know your thoughts.

Ken-Lo:
@megaslug Ah ok, never mind :) Didn't realise it was so difficult to re-use that existing logic, so this looks fine :) Thanks!

PR #812, opened 2016-12-30
Fix onClick not firing on mobile

megaslug:
CSS fix for `<Popover>` not triggering on mobile tap


clippy:

- [ ] Chrome
- [ ] IE11
- [ ] Firefox
- [ ] Safari
- [ ] iPad

PR #825, opened 2017-01-05
Remove affiliate tracking in splash pages

Ken-Lo:
We no longer need to set the `referral_code` cookie from within splash-pages, so now removing this code.


Ken-Lo:
Hey @megaslug 
Would you mind reviewing?
Thanks!

megaslug:
@Ken-Lo looks good!

PR #830, opened 2017-01-08
Fix layout in the lead data engineer job ad

hmac:
`className`s were all in the wrong place, so the bullet point lists looked a bit odd.

Before:
image

After:
image


hmac:
@megaslug could you review?

megaslug:
@hmac looks good!

PR #832, opened 2017-01-09
removal of partnership roles

katie-mai:
@megaslug - removal of both partnership roles. Please check and approve


megaslug:
@katie-mai since we're removing these two jobs, would be good to add redirects to both of these urls. See `redirects.json`

katie-mai:
@megaslug re redirecting - the are already commands to redirect on redirects.json > lines 4 and 12.

Shall we see if those work before deleting and re adding?

megaslug:
@katie-mai ok, that's cool. it's a bit worrying that they haven't worked

megaslug:
@katie-mai You removed `partnership-marketing-manager.js` and `partnerships-development.js` files, but these don’t match removals in `routes.js`, those are: `partnerships-development.js`, and `partnerships-success.js`

PR #837, opened 2017-01-10
Static site generation optimisation

megaslug:
As per @domderen [original PR](https://github.com/gocardless/splash-pages/pull/205) comment:

While preparing to exclude your this static page generation i thought of this small optimisation, and I thought you guys might be interested in it.

I am sending this PR mainly for the purposes of the peer review to see if I didn't miss any reasons why you used express in your static page generation process,
and to see if it could indeed be skipped, like I did here.

Thanks in advance for any feedback.

Static page rendering optimisation
- Bypassing express server in static site generation,
- dump-html now uses the same render function, as the express middleware,
- removed fetchPage method, as it was now not required,
- simplified build-prod script to only run production web pack build, and html-dump, no express sever in the middle.


megaslug:
hey @harrison, could you have a look at this PR? thanks

package.json
```diff
@@ -47,7 +47,6 @@
     "mime": "^1.3.4",
     "minimist": "^1.1.1",
     "mkdirp": "^0.5.1",
-    "nock": "^3.3.2",
```
harrison: update npm shrinkwrap? run `npm shrinkwrap` and hope it works.. happy to help on this, always a total pain. We should probably move to yarn so we don't have any issues with shrinkwrap..

script/dump-html
```diff
@@ -1,30 +1,32 @@
 #!/usr/bin/env babel-node
 
+
```
harrison: 🔪 newline

harrison:
@megaslug rebase against dev too?

server/render.js
```diff
@@ -107,7 +107,7 @@ export function renderFromRequest(req, res, next) {
   if (!isHtml) { return next(); }
 
   render(req.url)
-  .then(([, htmlOutput]) => {
+  .then((htmlOutput) => {
```
megaslug: @harrison it looks like removing [] from `.then(([htmlOutput]) => {` messes up the rendering of the page completely

harrison:
👍

